PR #6, opened 2019-05-04
10.0 calendar DST bug fix

luisiniguezh:
This fixes the known error on the calendar odoo module: https://github.com/odoo/odoo/issues/9268

Tests: 
To test this issue you can create a recurring event on the calendar before the DST change on your user timezone that spans to after the DST starts, you will notice that the recurring event is showing incorrectly after the date the DST starts, after the module is installed the recurring event is shown correctly.


calendar_dst_bug_fix/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "category": "Extra Tools",
+    "website": "https://github.com/OCA/calendar",
+    "author": "Luis Iñiguez, Odoo Community Association (OCA)",
+    "maintainers": ["luisiniguezh"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "preloadable": True,
+    "depends": [
+        "base",
```
luistorresm: This depends is not necessary, `calendar` already has it

calendar_dst_bug_fix/models/calendar_event.py
```diff
@@ -0,0 +1,47 @@
...
+        tz = self.env.context and self.env.context.get('tz')
+        if tz:
+            tz = pytz.timezone(tz)
+        for record in result:
+            id_pair = str(record['id']).split('-')
+            if (not record.get('allday')
+                    and isinstance(record['id'], (str))
+                    and tz and len(id_pair) > 1):
+                delta = self.get_time_dst_delta(int(id_pair[0]), tz, record)
+                if delta:
```
luistorresm:
```suggestion
                if not delta:
                    continue
```

To avoid many interations

calendar_dst_bug_fix/tests/test_calendar_recurrent_event_with_dst.py
```diff
@@ -0,0 +1,65 @@
...
+from odoo.tests import common
+from datetime import datetime
+import pytz
+
+
+class TestRecurrentEventWithDst(common.TransactionCase):
+
+    def setUp(self):
+        super(TestRecurrentEventWithDst, self).setUp()
+        self.CalendarEvent = self.env['calendar.event']
```
luistorresm:
```suggestion
        self.calendar_event = self.env['calendar.event']
```

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#variable-names

calendar_dst_bug_fix/models/calendar_event.py
luisiniguezh: @luistorresm that would mean to repeat the `data.append(record)` line right before the continue which doesn't seem DRY

luistorresm: This is to avoid the next iteration.

If you only need to update when delta is defined, then you could continue when is not present.

With this, line 27-31 will be moved 4 spaces before

luisiniguezh: the `data.append(record)` should be executed even if there's no delta so calling continue doesn't let that part of the code execute the code would have to look like:
```
...
                if not delta:
                    data.append(record)
                    continue
                for field in TIME_FIELDS_IN_EVENT:
                    if field in record and record[field]:
                        record[field] = api_fields.Datetime.to_string(
                            api_fields.Datetime.from_string(
                                record[field]) + delta)
            data.append(record)
...
``` 
is that preferred?

luistorresm: 👍

calendar_dst_bug_fix/readme/CREDITS.rst
```diff
@@ -0,0 +1,3 @@
+The development of this module has been financially supported by:
+
+* Advance Methods Co.
```
julioagf:
```suggestion
* Advanced Methods Co.
```

julioagf (approved):
Seems good

luistorresm (approved):

PR #13, opened 2020-04-08
[12.0] add calendar base booking

sebastienbeau:
Work in progress
The aim of this module is to add an abstract mixing for managing booking.
This can be added on every odoo model.
We build this dependency in order to had the feature of booking a datetime for retrieving the delivery into a physical shop (for Shopinvader), more module to come.

### DIFF
--- a/calendar_base_booking/__init__.py
+++ b/calendar_base_booking/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/calendar_base_booking/__manifest__.py
+++ b/calendar_base_booking/__manifest__.py
@@ -1,0 +1,18 @@
+{
+    "name": "Calendar Base Booking",
+    "summary": "Base module for adding booking feature",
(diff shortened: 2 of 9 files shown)


cubells:
hi @sebastienbeau 
if you need help on this give a ping.

sebastienbeau:
Thanks for proposing
I am going to rework on this next week and the week after (we deliverer in production the version 8 for the customer and new we can go back on v12 so it will be ready when we migrate the customer)
I plan to do some refactoring so we can reuse the feature in different context (renting, manufacturing...)
I will ping you for a review

sebastienbeau:
Hi just for information, I have some priority change.
So for now I will not work on it before september.
But it's still on my planning

sebastienbeau:
Closed in favours of https://github.com/OCA/calendar/pull/20 in V14

PR #14, opened 2020-08-11
[12.0] [MIG] calendar_resource

Rad0van:
Minimal changes required for it to work. Added calendar view for resource into menu. My motivation for this module was to use it as a mean for making reservations of meeting rooms so I wanted to add a view, where all meetings of that resources are visible. However in lower part responsible persons are shown and when selecting all this can override the selection and shows event not belonging to that resource. I guess some clever trick with domain is needed - any help/pointer would be appreciated.

@eilst, @max3903, @rimbendhaou and @hugho-ad would you please take a benevolent look? Thank you.


Rad0van:
So obviously it doesn't work as locally :-( Will try to fix that.

Rad0van:
So obviously I was developing this on Python 3.8 which seems to be a problem in multiple places. Installed 3.5, will try to fix this.

Rad0van:
Will redo completely. Closing now.

PR #22, opened 2021-01-20
[14.0][ADD] google_calendar_archive and google_calendar_hide_credentials

kevinkhao:
Add two minor modules: 
google_calendar_archive which makes the calendar view more usable (previously, users could not remove any calendar event that was synced, because only delete button is here and you can only archive synced calendar events)
google_calendar_hide_credentials simply hides google calendar credentials in the General Settings menu (before, you could see in plaintext everything just by opening that view)


mourad-ehm:
hi @kevinkhao, could you reopen this PR ?

PR #25, opened 2021-03-23
[ADD] resource_booking: new app to manage bookings

yajo:
This module adds a new app to allow you to book resource combinations in given
schedules.

Example use cases:

* Management of consultations in a clinic.
* Salesman appointments.
* Classroom and projector reservations.
* Hotel room booking.

Among the things you can do:

* Specify the type of booking, which includes a calendar of availability.
* Specify which resources can be booked together. All of them must be free to be booked.
* Place pending bookings, effectively giving permissions to someone to see the availability calendar and choose one slot.
* Partners can do that from their portals.
* If a partner has no user, he can still do the same via a tokenized URL.
* Backend users can also do that from the backend.
* Booking lifecycle with computed states.
* Automatic meeting creation and deletion.
* Automatic conflict detection.
* Deadline to block modifications.

@Tecnativa TT28201


sergio-teruel:
very red!! :smile:

yajo:
It should be less red now, thanks!

resource_booking/demo/res_users_demo.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2021 Tecnativa - Jairo Llopis
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<data>
+
+    <record id="base.user_demo" model="res.users">
+        <field name="groups_id" eval="[(4, ref('group_user'))]" />
+    </record>
```
joao-p-marques, line 9: Isn't this added by default?

resource_booking/models/calendar_event.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def write(self, vals):
+        """Check you're allowed to reschedule it."""
+        before = [(one.start, one.stop) for one in self]
+        result = super().write(vals)
+        rescheduled = self
+        for (old_start, old_stop), new in zip(before, self):
+            if old_start == new.start and old_stop == new.stop:
+                rescheduled -= new
+        rescheduled._validate_booking_modifications()
```
joao-p-marques, line 52: If this throws an exception, isn't it being thrown after the write in line 47?
If you are not allowed to reschedule it, shouldn't the error abort the write?

resource_booking/models/resource_booking_combination.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _get_intervals(self, start_dt, end_dt):
+        """Get available intervals for this booking combination."""
+        base = Intervals([(start_dt, end_dt, self)])
+        result = Intervals([])
+        for combination in self:
+            combination_intervals = base
+            for res in combination.resource_ids:
+                if not combination_intervals:
+                    break  # Can't restrict more
```
joao-p-marques, line 78: Why putting this check inside the loop? You could avoid entering that `if` in every iteration by:
```suggestion
        for combination in self:
            combination_intervals = base
            if combination_intervals:
                for res in combination.resource_ids:
```

resource_booking/views/resource_booking_views.xml
```diff
@@ -0,0 +1,130 @@
...
+    <!-- Views -->
+    <record id="resource_booking_view_calendar" model="ir.ui.view">
+        <field name="name">Resource booking calendar</field>
+        <field name="model">resource.booking</field>
+        <field name="arch" type="xml">
+            <calendar string="Bookings" date_start="start" date_stop="stop" quick_add="false" color="combination_id">
+                <field name="name" />
+                <field name="partner_id" avatar_field="image_small" />
+                <field name="combination_id" />
+                <!-- <field name="is_highlighted" invisible="1" /> -->
```
joao-p-marques: Should this be removed?

joao-p-marques (commented):
Code and functional review :+1: 

Just some tiny comments. :smiley: But overall LGTM

resource_booking/demo/res_users_demo.xml
yajo: Nope, I'm making "Marc Demo" a resource booking user. Notice the ref is `group_user` and not `base.group_user`. When not indicating the module, it assumes the current one and translates into `resource_booking.group_user`.

resource_booking/models/calendar_event.py
yajo: Yes, the transaction will be rolled back, so the final behavior is the one you say.

resource_booking/models/resource_booking_combination.py
yajo: Yes, but `combination_intervals` gets more restricted with each loop (see next lines). If, after some iteration, it becomes void, we don't need to continue checking.

yajo:
All attended/responded.

resource_booking/demo/res_users_demo.xml
joao-p-marques: Oh ok, of course. In that case, maybe a more specific name to the group wouldn't harm, but that is already nitpicking, not important :+1:

resource_booking/models/resource_booking_combination.py
joao-p-marques: Ok, thanks, I hadn't seen that `combination_intervals` changed bellow...

joao-p-marques (approved):
> All attended/responded.

Ok, thanks for the pointers :+1: 

Looks good now

sergio-teruel (approved):

PR #27, opened 2021-05-06
[FIX] resource_calendar: ignore past or unconfirmed bookings when updating calendars

yajo:
Without this patch, users couldn't change a calendar schedule if there were past or unconfirmed bookings that wouldn't fit in it.

Excluding those bookings from the check fixes the situation.

We also check that, to confirm a booking, it must fit in the calendar (because now it can happen that, in the time that has passed since the booking was scheduled until it is confirmed, the calendar changes).

@Tecnativa TT29509

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -188,6 +188,7 @@ class ResourceBooking(models.Model):
     @api.depends("active", "meeting_id.attendee_ids.state")
     def _compute_state(self):
         """Obtain request state."""
+        to_check = self.browse(prefetch=self._prefetch)
         for one in self:
             if not one.active:
                 one.state = "canceled"
@@ -199,8 +200,10 @@ class ResourceBooking(models.Model):
                     break
             if confirmed:
                 one.state = "confirmed"
+                to_check |= one
                 continue
             one.state = "scheduled" if one.meeting_id else "pending"
+        to_check._check_scheduling()
 
     @api.depends("meeting_id.start", "meeting_id.stop")
     def _compute_dates(self):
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I don't get the use case if this leads to something forbidden...

victoralmau (approved):

yajo:
It was taking into consideration past bookings too. Do you think it's better to narrow the search?

pedrobaeza:
Although past events, if they were booked, why now ignore them? Shouldn't they be removed first?

yajo:
No, there's no reason to remove past bookings.

Just we shouldn't forbid altering the calendar just because some booking in the past didn't fit in it. The calendar is only used to allow bookings, but you can't book something for the past, only for the future.

pedrobaeza:
OK, then forbid only for booking on the future, yes

carlosdauden (approved):

yajo:
OK, I changed the implementation and instead ignored past and unconfirmed events. I hide previous conversation and changed description and title, as this is completely different. Please review again. Thanks everyone.

pedrobaeza:
I'm still not sure about the unconfirmed part, as they can conflict when confirming them.

yajo:
Yes, that's the goal.

Imagine your customer prereserved some slot, but then you change the calendar, and later the customer tries to confirm. IMHO:

- the system shouldn't forbid you altering the calendar just because your customer didn't finish the process.
- the customer should get an error if confirming late, and thus he should reschedule.

victoralmau (approved):

pedrobaeza:
But such error confirming the booking exists?

yajo:
Yes, but let me add that assertion to the test.

yajo:
Oops, I was wrong. Indeed the test showed that it could happen. I added the fix and the test.

chienandalu (approved):

pedrobaeza (approved):

PR #28, opened 2021-06-08
[FIX] resource_booking: always notify in resource TZ

yajo:
The notifications emitted to the resource booking requester must always be in the same TZ as the resource booking itself.

For example, if you book one hotel room in the other side of the world, a notification in your own TZ is confusing.

Besides, res.partner created from website_sale are created with `tz=False`, making it even more confusing.

@Tecnativa TT30331

### DIFF
--- a/resource_booking/models/calendar_event.py
+++ b/resource_booking/models/calendar_event.py
@@ -51,3 +51,13 @@ class CalendarEvent(models.Model):
                 rescheduled -= new
         rescheduled._validate_booking_modifications()
         return result
+
+    def get_interval(self, interval, tz=None):
+        """Autofix tz from related resource booking.
+
+        This function is called to render calendar.event notification mails.
+        Any notification related to a resource.booking must be emitted in the
+        same TZ as the resource.booking. Otherwise it's confusing to the user.
+        """
+        tz = self.resource_booking_ids.type_id.resource_calendar_id.tz or tz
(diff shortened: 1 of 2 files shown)


resource_booking/models/calendar_event.py
```diff
@@ -51,3 +51,13 @@ def write(self, vals):
...
+
+    def get_interval(self, interval, tz=None):
+        """Autofix tz from related resource booking.
+
+        This function is called to render calendar.event notification mails.
+        Any notification related to a resource.booking must be emitted in the
+        same TZ as the resource.booking. Otherwise it's confusing to the user.
+        """
+        tz = self.resource_booking_ids.type_id.resource_calendar_id.tz or tz
+        return super().get_interval(interval, tz)
```
pedrobaeza: Keep keyword arguments as keyword, not positional

pedrobaeza (approved):

hparfr (approved):

joao-p-marques (approved):
Code review :+1:

resource_booking/models/calendar_event.py
yajo: Thanks for the suggestion, but I don't see any advantages.

pedrobaeza: It's not a suggestion, it's a Python guideline.

yajo: Done.

PR #34, opened 2021-07-22
[IMP+FIX] resource_booking: duration, location, involves me, hr_holidays_public

yajo:
This PR includes several improvements or fixes for the `resource_booking` module:

##  [IMP] resource_booking: use duration field instead of stop 

From now on, `resource.booking.type` duration represents the default duration that will be applied to new `resource.booking` of that type, instead of the hardcoded duration for every new booking.

This means that you can create a new booking in pending state and define a custom duration before sending the portal link to customer.

Thus, the available schedule slots in portal will be based on the type duration, whereas only those that have enough time left to complete the booking duration will be displayed.

image

##  [IMP] resource_booking: configurable auto-assigned combination and location

Thanks to this patch, you will be able to more predictably block the chosen resource combination, with the added checkbox.

Also, the location field is propagated to `resource.booking` records, and synced from there to `calendar.event` if needed. This way, you can alter the location before sending the portal invitation.

image

##  [FIX] resource_booking: "Involves me" filter 

The filter wasn't working fine due to the `auto_join=True` set in `meeting_id`. It was never displaying pending bookings.

Besides, it wasn't necessary to keep it as a field because it was displayed nowhere and the search could be done directly in the view. This way, there's less python code to maintain and we disable the possibility of having a negative domain, which enters doomed scenarios when x2many fields are involved (see odoo/odoo#43957).

##  [IMP] resource_booking: compatible with hr_holidays_public 

Adding `.with_context(exclude_public_holidays=True)` in needed places to let `hr_holidays_public` do its magic and never allow allocating a booking during a public holidays when that module is properly installed and configured.

image

## [IMP] resource_booking: optional name propagated to meeting

- Allow users to optionally define a name for the resource booking.
- Display that name in portal if set.
- Sync it with meeting name when creating it.

https://user-images.githubusercontent.com/973709/127126593-701f0fe7-1afb-41aa-a7c8-2f869a89e563.mp4

@Tecnativa TT30987

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -61,6 +61,14 @@ class ResourceBooking(models.Model):
         store=True,
         tracking=True,
     )
+    combination_auto_assign = fields.Boolean(
+        string="Auto assigned",
+        default=True,
+        help=(
+            "When checked, resource combinations will be (un)assigned automatically "
+            "based on their availability during the booking dates."
+        ),
+    )
     name = fields.Char(compute="_compute_name")
     partner_id = fields.Many2one(
         "res.partner",
@@ -71,6 +79,7 @@ class ResourceBooking(models.Model):
         tracking=True,
         help="Who requested this booking?",
     )
+    location = fields.Char(compute="_compute_location", readonly=False, store=True,)
     requester_advice = fields.Text(related="type_id.requester_advice", readonly=True)
     involves_me = fields.Boolean(
         compute="_compute_involves_me", search="_search_involves_me"
@@ -100,14 +109,13 @@ class ResourceBooking(models.Model):
         compute="_compute_start",
         copy=False,
         index=True,
-        inverse="_inverse_dates",
+        readonly=False,
         store=True,
         track_sequence=200,
         tracking=True,
     )
     duration = fields.Float(
         compute="_compute_duration",
-        inverse="_inverse_dates",
         readonly=False,
         store=True,
         track_sequence=220,
@@ -144,12 +152,12 @@ class ResourceBooking(models.Model):
             if one.type_id:
                 one.categ_ids = one.type_id.categ_ids
 
-    @api.depends("start", "stop", "type_id")
+    @api.depends("start", "type_id", "combination_auto_assign")
     def _compute_combination_id(self):
         """Select best combination candidate when changing booking dates."""
         for one in self:
             # Useless without the interval
-            if one.start and one.stop:
+            if one.start and one.combination_auto_assign:
                 one.combination_id = one._get_best_combination()
 
     @api.depends("combination_id", "partner_id")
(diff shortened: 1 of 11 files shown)


resource_booking/models/resource_booking.py
```diff
@@ -231,31 +230,53 @@ def _compute_state(self):
...
         for one in self:
-            # You're creating a new record; calendar view sends proper context
-            # defaults that at this point are lost; restoring them
+            if one.id:
+                one.start = one.meeting_id.start
+
+    @api.depends("meeting_id.duration", "type_id")
+    def _compute_duration(self):
+        """Compute duration for each booking."""
+        for one in self:
```
pedrobaeza: I don't like too much this notation of `one` and prefer `record`, that is more standard.

pedrobaeza (changes requested):
On a functional review, everything goes correct, but the booking confirmation shows the incorrect duration:

Selección_014

yajo:
All done. Please keep in mind that this will have to be merged with "nobump", as it includes the version bump needed to test migration scripts.

Yajo:
Added new IMP to the PR:

> Thanks to this patch, you will be able to more predictably block the chosen resource combination, with the added checkbox.
>
> Also, the location field is propagated to `resource.booking` records, and synced from there to `calendar.event` if needed. This way, you can alter the location before sending the portal invitation.

yajo:
Updated description to reflect current state of the PR, which has 4 different feature sets. Ready to review/merge.

resource_booking/models/resource_booking.py
```diff
@@ -100,14 +109,13 @@ class ResourceBooking(models.Model):
         compute="_compute_start",
         copy=False,
         index=True,
-        inverse="_inverse_dates",
+        readonly=False,
```
pedrobaeza, line 107: Haven't you mixed these changes in an unrelated commit?

pedrobaeza (commented):
This is still missing the `name` field on the resource.booking, needed for setting the meeting topic and to be transferred to the calendar.event.

resource_booking/models/resource_booking.py
yajo: AFAICS it's OK. I'm removing the inverse, but still allowing writes with readonly=False.

yajo:
> This is still missing the `name` field on the resource.booking, needed for setting the meeting topic and to be transferred to the calendar.event.

Currently the name is computed based on the date, requester and type, and it's propagated to the calendar meeting.

That's on purpose because originally having to add a name would be uncomfortable and unnecessary, because basically every booking of the same type has the same purpose. Also, when placing a booking from e-commerce, we don't want the customer to have to write a name; that adds no value.

I can add the `name` field and make it optional. Then, if it's empty, I can compute the `display_name` to the same value `name` has now. Sounds good?

resource_booking/models/resource_booking.py
pedrobaeza: I'm saying that the changes are on the second commit, while the topic belongs to the first commit.

pedrobaeza:
I want to add a free text. For example: "Meeting about module X". And the title of the calendar.event should be that one, and in the portal that should be also. You tell me the technical specifications for that.

yajo:
Ok, but I mean that we need to make that optional. We can't require the name field to be filled, or we'll break the original purpose of the module.

pedrobaeza:
It can be optional indeed.

resource_booking/models/resource_booking.py
yajo: I double-checked it and it's OK. The need to change from `_inverse_dates` to `_sync_dates` happened in 1cd849d8c40f644aefb62391ce0c3dfe6dd78012.

yajo:
Custom optional name added (see last video on description).

pedrobaeza (approved):

sergio-teruel (approved):

PR #37, opened 2021-07-29
[IMP] resource_booking: autoconfirm preselected combination resource attendees

yajo:
When a booking has a preselected resource combination, the resource attendees don't need to reconfirm their attendance to the meeting, because they've already stated that they'll be available on those dates.

@Tecnativa TT31239

### DIFF
--- a/resource_booking/models/calendar_event.py
+++ b/resource_booking/models/calendar_event.py
@@ -52,6 +52,42 @@ class CalendarEvent(models.Model):
         rescheduled._validate_booking_modifications()
         return result
 
+    def create_attendees(self):
+        """Autoconfirm resource attendees if preselected."""
+        old_attendees = self.attendee_ids
+        result = super(
+            # This context avoids sending invitations to new attendees
+            CalendarEvent,
+            self.with_context(detaching=True),
+        ).create_attendees()
+        new_attendees = (self.attendee_ids - old_attendees).with_context(
+            self.env.context
+        )
+        for attendee in new_attendees:
+            # No need to change state if it's already done
+            if attendee.state in {"accepted", "declined"}:
+                continue
+            rb = attendee.event_id.resource_booking_ids
+            # Confirm requester attendee always if requested
+            if (
+                self.env.context.get("autoconfirm_booking_requester")
+                and attendee.partner_id == rb.partner_id
+            ):
+                attendee.state = "accepted"
+                continue
+            # Auto-confirm if attendee comes from a handpicked combination
+            if rb.combination_auto_assign:
+                continue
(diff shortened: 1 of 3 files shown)


resource_booking/models/calendar_event.py
```diff
@@ -52,6 +52,24 @@ def write(self, vals):
         rescheduled._validate_booking_modifications()
         return result
 
+    def create_attendees(self):
+        """Autoconfirm resource attendees if preselected."""
+        old_attendees = self.attendee_ids
```
pedrobaeza, line 57: Are they going to be any?

resource_booking/models/calendar_event.py
```diff
@@ -52,6 +52,24 @@ def write(self, vals):
...
+        result = super().create_attendees()
+        new_attendees = self.attendee_ids - old_attendees
+        for attendee in new_attendees:
+            # No need to change state if it's already done
+            if attendee.state in {"accepted", "declined"}:
```
pedrobaeza, line 68: Should we use `filtered` in the loop for avoiding 2 lines of code?

yajo: When using `.filtered()`, Odoo needs to loop over all records in the set. Then, after building a new set without the discarded ones, it would have to loop over them again because of the `for` sentence. Also, lambdas tend to be a bit less readable and harder to debug in certain circumnstances.

For these reasons, these days I'm changing my habit and, when I need to loop and discard elements, I just discard them within the loop. It seems clearer and it will be more performant.

Still, I use `filtered` when I really need a recordset for doing for example bitwise operations.

resource_booking/models/calendar_event.py
yajo: Yes, it's possible according to upstream logic.

resource_booking/models/calendar_event.py
pedrobaeza: The performance reasoning is not true: cache makes the "double" loop to not impact. About the readability, I prefer one line with the criteria that 3. Lastly, with the `filtered` option, you don't have uncovered lines by tests.

yajo: > Lastly, with the `filtered` option, you don't have uncovered lines by tests

That'd be a false positive.

pedrobaeza: I don't think so, but OK, we will do this differently when coding our PRs.

pedrobaeza (approved):

chienandalu (approved):

CarlosRoca13 (approved):

sergio-teruel (approved):

PR #38, opened 2021-07-29
[FIX] resource_booking: allow combination with other kind of leaves

yajo:
When `_leave_intervals_batch()` in `resource.calendar` is extended by other modules that add more leaves for any reason, the returned recordset must be of the `resource.calendar.leaves` model. Otherwise, Odoo fails with:

    TypeError: '<' not supported between instances of 'resource.resource' and 'resource.calendar.leaves'

The resource didn't have a meaning here, so changing it to avoid bug.

@Tecnativa TT31249

### DIFF
--- a/resource_booking/models/resource_calendar.py
+++ b/resource_booking/models/resource_calendar.py
@@ -89,7 +89,7 @@ class ResourceCalendar(models.Model):
                         fields.Datetime.context_timestamp(
                             event, fields.Datetime.to_datetime(event.stop)
                         ),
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

CarlosRoca13 (approved):

pedrobaeza:
Fast-tracking as bugfix with tests:

/ocabot merge patch

chienandalu (approved):

PR #42, opened 2021-08-02
[FIX] resource_booking: suggest involved partners as followers

yajo:
Includes those coming from requester and from resource combination.

Fixes this:

pantallazo_2021-09-15_13-15

@Tecnativa TT31901

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -532,12 +532,20 @@ class ResourceBooking(models.Model):
             new.append((id_, name))
         return new
 
-    def message_get_suggested_recipients(self):
-        recipients = super().message_get_suggested_recipients()
-        for one in self:
-            if one.partner_id:
-                one._message_add_suggested_recipient(
-                    recipients, partner=one.partner_id, reason=_("Requesting partner")
+    def _message_get_suggested_recipients(self):
+        """Suggest related partners."""
+        recipients = super()._message_get_suggested_recipients()
+        for record in self:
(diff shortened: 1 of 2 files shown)


resource_booking/models/resource_booking.py
```diff
@@ -377,6 +377,23 @@ def _check_scheduling(self):
                 % "\n- ".join(unfitting_bookings.mapped("display_name"))
             )
 
+    def _message_get_suggested_recipients(self):
+        """Suggest related partners."""
+        result = super()._message_get_suggested_recipients()
+        for record in self:
+            record._message_add_suggested_recipient(
```
pedrobaeza: Why are you calling directly to the method? Please follow inheritance compatibility adding the ones you want to the `result` variable.

yajo: [The method docstring](https://github.com/odoo/odoo/blob/cc6c61d4c88aa9dfae2c343356da249d1f22b5f4/addons/mail/models/mail_thread.py#L1521-L1523) says:

> Called by _message_get_suggested_recipients, to add a suggested recipient in the result dictionary.

So this is the proper method of doing that.

pedrobaeza: True. Core does that: https://github.com/odoo/odoo/blob/91c55ab8393c47b18be31f1f5cac74b3673cdf4b/addons/project/models/project.py#L1308

It seems a bit convoluted, but it's the way to do it.

pedrobaeza (approved):

PR #45, opened 2021-09-27
[FIX] resource_booking: autosubscribe resource partners

yajo:
When some partner is a resource assigned to a booking, he should get auto-subscribed.

Inspired in https://github.com/OCA/calendar/pull/42, but I'm changing that behavior.

@Tecnativa TT32148

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -446,12 +446,35 @@ class ResourceBooking(models.Model):
         result &= combinations._get_intervals(start_dt, end_dt)
         return result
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

joao-p-marques (approved):
Code review :+1:

pedrobaeza:
Does some of this be forward-ported?

joao-p-marques:
I believe so, but @Yajo is already on it...

yajo:
Yes, will do after fixing a regression introduced.

PR #46, opened 2021-09-28
[FIX] resource_booking: correct assignment message

yajo:
By using the upstream `mail.message_user_assigned` template, we were getting this exception:

```
odoo.addons.base.models.qweb.QWebException: 'resource.booking' object has no attribute 'user_id'
Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/qweb.py", line 348, in _compiled_fn
    return compiled(self, append, new, options, log)
  File "<template>", line 1, in template_2601_10519
AttributeError: 'resource.booking' object has no attribute 'user_id'
```

So I provide a specific template for this kind of subscription notification, which doesn't use that `user_id` field. Besides, it explains better why you're getting it.

@Tecnativa TT32190

### DIFF
--- a/resource_booking/__manifest__.py
+++ b/resource_booking/__manifest__.py
@@ -27,6 +27,7 @@
         "web_calendar_slot_duration",
     ],
     "data": [
+        "data/mail.xml",
         "security/resource_booking_security.xml",
         "security/ir.model.access.csv",
         "templates/assets.xml",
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Isn't better to have a field `user_id` (computed or something) than all of this stuff?

yajo:
That field was added in v13, but it has a different purpose.

One thing is that you're the organizer of the booking (user_id), but another thing is that you are the booked resource (this case).

As both things are different and the one being requested here is the latter, I'm not backporting any of the v13 stuff here. But I'll fwport this patch.

PR #47, opened 2021-09-28
[FIX] resource_booking: allow combination with other kind of leaves

yajo:
When `_leave_intervals_batch()` in `resource.calendar` is extended by other modules that add more leaves for any reason, the returned recordset must be of the `resource.calendar.leaves` model. Otherwise, Odoo fails with:

    TypeError: '<' not supported between instances of 'resource.resource' and 'resource.calendar.leaves'

The resource didn't have a meaning here, so changing it to avoid bug.

@Tecnativa TT31249 TT32190 - backport of https://github.com/OCA/calendar/pull/38

### DIFF
--- a/resource_booking/models/resource_calendar.py
+++ b/resource_booking/models/resource_calendar.py
@@ -83,7 +83,7 @@ class ResourceCalendar(models.Model):
                     (
                         fields.Datetime.context_timestamp(event, event.start),
                         fields.Datetime.context_timestamp(event, event.stop),
-                        event,
+                        self.env["resource.calendar.leaves"],
                     )
                 )
         return Intervals(intervals)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Isn't this a problem of `hr_holidays_public` instead? We fixed a similar thing in 13.0 cc @victoralmau

pedrobaeza (approved):
I see that the problem is here, but the solution was the same as we did in `hr_holidays_public`

pedrobaeza:
Please check Travis. Probably the test was not correct (although the fix is). Didn't you run it locally?

yajo:
It was my fault on a conflict resolution... I introduced a syntax error. Fixed now.

/ocabot merge patch

yajo:
Ah wait, there's also a bug in the test as you said. I cancelled the merge.

yajo:
Finally, the good one!

/ocabot merge patch

PR #55, opened 2021-10-26
[FIX] resource_booking: free meetings by default

yajo:
If I'm creating bookings for my colleagues, I'll be the owner of all of the meetings. But that doesn't mean I'll be attending. And OTOH, if I'm part of the bookied resource combination, I'll already be busy due to my attendance status.

So it makes sense to be free by default. However, if I still choose to be busy at that time, I'll create an unavailability as expected.

@Tecnativa TT32666

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -238,6 +238,12 @@ class ResourceBooking(models.Model):
                     resource_booking_ids=[(6, 0, one.ids)],
                     start=one.start,
                     stop=one.stop,
+                    # If you're not booked, you're free
+                    show_as=(
+                        "busy"
+                        if self.env.user.partner_id in resource_partners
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

pedrobaeza:
So if I'm part of the booked resources, it will appear as busy?

yajo:
No, but you'll be a confirmed attendee, so your slot will still be unavailable.

pedrobaeza:
But I want that in my agenda it appears as busy, that has nothing to do with the resource booking system.

pedrobaeza:
Can't you do this only if the owner is not attending?

yajo:
Makes sense, changed.

yajo:
All attended, so merging.

/ocabot merge patch

PR #56, opened 2021-10-26
[FIX] resource_booking: free meetings by default

yajo:
If I'm creating bookings for my colleagues, I'll be the owner of all of the meetings. But that doesn't mean I'll be busy.

With this patch, only when I'm a booked resource will the event be set as busy.

@Tecnativa TT32666 Fw port of https://github.com/OCA/calendar/pull/55


yajo:
Fast tracking fwport

/ocabot merge patch

PR #65, opened 2022-03-14
[13.0][FIX] resource_booking: calendar view performance

chienandalu:
We need a stop parameter. Otherwise, the calendar view could be trying
to render tons of records making the client's browser have a hard time.

cc @Tecnativa TT35062

please check @pedrobaeza @sergio-teruel


pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
Let's merge it as it's critical for performance:

/ocabot merge patch

PR #68, opened 2022-06-08
[13.0][FIX] resource_booking: performance with recurrent events

chienandalu:
`rrule` isn't very performant on iterations. This is used on calendar
events to compute the virtual events set.

We want to avoid calling it too much.

Real case with website_sale_resource_booking (thousands of virtual events):

before: /shop/booking/<id>/schedule: ~25 seconds
after: /shop/booking/<id>/schedule: ~6 seconds

before: /shop/booking/<id>/confirm: ~20 seconds
after: /shop/booking/<id>/confirm: ~3 seconds

before: /shop/checkout: ~18 seconds
after: /shop/checkout: ~2 seconds

before: /shop/address: ~18 seconds
after: /shop/address: ~2 seconds

There's room for improvement anyway

An speed scope for /shop/address before and after:

before:
image

after:
image

cc @Tecnativa TT37089

ping @pedrobaeza @sergio-teruel @carlosdauden

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -505,7 +505,10 @@ class ResourceBooking(models.Model):
 
     def write(self, vals):
         """Sync booking with meeting if needed."""
-        result = super().write(vals)
+        # On a database with lots of recurrent calendar events we could get serious
+        # performance downgrades. As we'll be computing them with _sync_meeting later
+        # we'll be avoiding triggering on the super call (i.e. compute methods)
+        result = super(ResourceBooking, self.with_context(virtual_id=False)).write(vals)
         self._sync_meeting()
         return result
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Please check pre commit

chienandalu:
> Please check pre commit

Maybe it's needed to update the branch CIs?

yajo:
Have you tested with https://github.com/odoo/odoo/pull/69822?

chienandalu:
> Have you tested with [odoo/odoo#69822](https://github.com/odoo/odoo/pull/69822)?

Thanks @yajo Yes, I checked it but I wanted to avoid lateral issues from such a big change. And the performance gain isn't that much added to this one, I'm afraid (although there's gain). We'll try to get along with these performance times until the next version and keep recurrent events as low as possible.

yajo (approved):

PR #70, opened 2022-06-16
[FIX] calendar_partner_color display colors on filters

remi-filament:
When using this module, the colors are properly displayed on calendar, although there is no legend anymore, meaning that on the right section where you can filter on what to display, you do not get any color on owner, so it is very difficult to assess which color belongs to whom...

I added dependency on web_calendar_color_field (from OCA web repo) in order to display the color on filters as well.


amh-mw:
This certainly makes me wonder why `web_calendar_color_field` isn't in this repository as `calendar_color_field`?

thomaspaulb (approved):

gurneyalex (approved):

PR #73, opened 2022-10-24
[14.0][MIG] resource_booking: Migration to 14.0

pedrobaeza:
- Standard procedure
- No more virtual ids
- Adapt tests
- FIX: Proper show_as meeting generation
- IMP: Add name to tree view

@Tecnativa TT32344


pedrobaeza:
@njeudy ready to be reviewed

resource_booking/__manifest__.py
```diff
@@ -1,24 +1,23 @@
...
     "summary": "Manage appointments and resource booking",
-    "version": "13.0.2.6.5",
-    "development_status": "Beta",
+    "version": "14.0.1.0.0",
+    "development_status": "Production/Stable",
     "category": "Appointments",
     "website": "https://github.com/OCA/calendar",
     "author": "Tecnativa, Odoo Community Association (OCA)",
-    "maintainers": ["Yajo"],
+    "maintainers": ["pedrobaeza"],
```
yajo, line 13: Thanks for taking maintainership of this module. ❤️ In the future I'll appreciate it if you ping me about these kind of changes instead of just doing them without notice.

pedrobaeza: Well, you are not anymore with us, so it's logic that you are not going to maintain this module. But if you prefer to keep maintaining it, I'm glad to leave it to you. But you can't retain something that you are not going to maintain, and you haven't expressed such desire nor do anything in this module since your departure, so I don't feel in the obligation of warning your for each module you have done under Tecnativa.

resource_booking/models/resource_booking.py
```diff
@@ -304,12 +302,7 @@ def _sync_meeting(self):
                     start=one.start,
                     stop=one.stop,
                     user_id=one.user_id.id,
-                    # If you're not booked, you're free
-                    show_as=(
-                        "busy"
-                        if self.env.user.partner_id in resource_partners
-                        else "free"
-                    ),
+                    show_as="busy",
```
huguesdk, line 305: why is this change removed here (and the test for the `show_as` property value in `test_backend.BackendCase.test_event_show_as_free()`), while it was added [recently in 13.0](https://github.com/OCA/calendar/commit/9e32cf5ddf421b341c1ee5de7b1fab9939d76c6d)? if the lines checking for the value are re-added to the test, it fails.

edit: it was not added recently to 13.0, this was exactly one year ago. ☺️ but the question remains: why remove it in 14.0?

pedrobaeza: Because this was a misconception: the meeting should always be busy, as the current user shouldn't influence on the state. Do you have any use case about this?

huguesdk: no, no use case, only wondering about the reason of the change. thanks for the explanation!

pedrobaeza: The misconception is that one: other colleague creating resource bookings is not going to be on the attendee list if they simply aren't in the resource combination list nor organizer.

gurneyalex (approved):

PR #77, opened 2023-01-31
[14.0][IMP] resource_booking: Allow a booking to span more than one calendar day

carmenbianca:
This PR solves two issues:

- Can't create a booking that lasts from e.g. 23:00 to 1:00.
- Can't create a booking that lasts several days (e.g. rentals).

Internal task: https://gestion.coopiteasy.be/web#id=9834&action=475&active_id=434&model=project.task&view_type=form&menu_id=536

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -14,12 +14,61 @@ from odoo.addons.resource.models.resource import Intervals
 
 
 def _availability_is_fitting(available_intervals, start_dt, end_dt):
+    # Test whether the stretch between start_dt and end_dt is an uninterrupted
+    # stretch of time as determined by `available_intervals`.
+    #
+    # `available_intervals` is typically created by `_get_intervals()`, which in
+    # turn uses `calendar._work_intervals()`. It appears to be default upstream
+    # behaviour of `_work_intervals()` to create a (start_dt, end_dt, record)
+    # tuple for every day, where end_dt is at 23:59, and the next tuple's
+    # start_dt is at 00:00.
+    #
+    # Changing this upstream behaviour of `_work_intervals()` to return a
+    # _single_ tuple for any multi-day uninterrupted stretch of time would
+    # probably be preferable, but (1.) the code in `_work_intervals()` is
+    # unbelievably arcane, and (2.) changing this behaviour is extremely likely
+    # to cause bugs elsewhere. So instead, we account for the upstream behaviour
+    # here.
+    start_date = start_dt.date()
+    end_date = end_dt.date()
     # Booking is uninterrupted on the same calendar day.
-    return (
+    if (
         len(available_intervals) == 1
         and available_intervals._items[0][0] <= start_dt
         and available_intervals._items[0][1] >= end_dt
-    )
+    ):
+        return True
+    # Booking spans more than one calendar day, e.g. from 23:00 to 1:00
+    # the next day.
+    elif available_intervals and start_date != end_date:
+        tally_date = start_date
+        for item in available_intervals:
+            item0_date = item[0].date()
+            item1_date = item[1].date()
+            # FIXME: Really weird workaround for when available_intervals has
+            # nonsensical items in it where item1_date is before item0_date.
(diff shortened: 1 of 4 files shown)


carmenbianca:
Ready for review after a fair amount of tinkering.

carmenbianca:
@huguesdk I tried hacking the same result by forcibly setting `hour_to` to 24.0 on `resource.calendar.attendance` records -> didn't work.

resource_booking/models/resource_booking.py
```diff
@@ -13,6 +13,38 @@
...
+        end_dt,
+    ):
+        return True
+    # Booking spans more than one calendar day, e.g. from 23:00 to 1:00
+    # the next day.
+    elif (
+        available_intervals
+        and start_date != end_date
+        and (end_date - start_date).days == (len(available_intervals) - 1)
+    ):
```
carmenbianca: We could improve this check by removing it, and iterating over every item to check whether they are on subsequent dates, keeping a rolling tally.

carmenbianca: implemented

resource_booking/models/resource_booking.py
```diff
@@ -13,6 +13,44 @@
...
+            item0_date = item[0].date()
+            item1_date = item[1].date()
+            # FIXME: Really weird workaround for when available_intervals has
+            # nonsensical items in it where item1_date is before item0_date.
+            # Just ignore those items and pretend they don't exist; all the
+            # other items appear to make sense.
+            if item1_date < item0_date:
+                continue
+            # Intervals that aren't on the running tally date break the streak.
+            if item0_date != tally_date or item1_date != tally_date:
```
huguesdk, line 57: i find this a little confusing. i first thought that there was an error, because this means that `item0_date` and `item1_date` must both be equal to `tally_date`, and i was wondering how the interval could span multiple days in this case. but here we are iterating over items that span hours on the same date, so it works.

in which case is this condition true? adding a test for this would be useful.

resource_booking/models/resource_booking.py
```diff
@@ -13,6 +13,44 @@
...
+                break
+            # Intervals that aren't on the end date should end at 23:59.
```
huguesdk:
```suggestion
            # Intervals that aren't on the end date should end at 23:59 (and any number of seconds).
```

resource_booking/models/resource_booking.py
```diff
@@ -13,6 +13,44 @@
...
+        tally_date = start_date
+        for item in available_intervals:
+            item0_date = item[0].date()
+            item1_date = item[1].date()
+            # FIXME: Really weird workaround for when available_intervals has
+            # nonsensical items in it where item1_date is before item0_date.
+            # Just ignore those items and pretend they don't exist; all the
+            # other items appear to make sense.
+            if item1_date < item0_date:
+                continue
```
huguesdk, line 53: adding a test for this would be useful.

huguesdk (commented):
thanks for this. it’s a little sad that this is not handled correctly by the core `resource` module. adding a comment explaining how this is a workaround for those limitations would be useful (`resource.calendar.attendance` cannot end at midnight so cannot end exactly at the end of the day (only a few milliseconds before)).

resource_booking/models/resource_booking.py
carmenbianca: added a test.

you're right. documented this better.

carmenbianca:
This is now rebased on top of #80. Will need to merge that PR before merging this.

carmenbianca:
Now that #80 is merged, I rebased this on top of 14.0.

remytms (approved):
LGTM

norlinhenrik:
LGTM
When I configure the resource calender with work hours every day 00:00-23:59, I can book a resource for a month. :)

pedrobaeza:
Please squash commits a bit and check pre-commit.

carmenbianca:
Rebased on top of 14.0. Commits squashed. pre-commit works locally.

pedrobaeza:
/ocabot merge patch

You can now fw-port it to upper versions.

PR #78, opened 2023-02-02
[ADD][14.0] resource_booking_simultaneous_human

carmenbianca:
When a resource of type human is scheduled for a booking, they remain
simultaneously available for other bookings.

The purpose of this module is bookings for events where a worker can split their
attention between multiple attendants.

Internal task: https://gestion.coopiteasy.be/web#id=9879&action=475&active_id=434&model=project.task&view_type=form&menu_id=536

### DIFF
--- a/resource_booking_simultaneous_human/__manifest__.py
+++ b/resource_booking_simultaneous_human/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2023 Coop IT Easy SC
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Resource booking simultaneous availability for humans",
+    "summary": """
+        When a resource of type human is scheduled for a booking, they will remain
+        simultaneously available for other bookings.""",
+    "version": "14.0.1.0.0",
+    "category": "Appointments",
+    "website": "https://github.com/OCA/calendar",
+    "author": "Odoo Community Association (OCA), Coop IT Easy SC",
+    "maintainers": ["carmenbianca"],
+    "license": "AGPL-3",
+    "depends": [
+        "resource_booking",
+    ],
+}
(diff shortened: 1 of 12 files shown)


carmenbianca:
I think the broken pre-commit test is unrelated to these changes.

resource_booking_simultaneous_human/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "maintainers": ["carmenbianca"],
+    "license": "AGPL-3",
+    "application": False,
+    "depends": [
+        "resource_booking",
+    ],
+    "excludes": [],
+    "data": [],
+    "demo": [],
+    "qweb": [],
```
huguesdk: avoid empty keys (see [here](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules)). i would also remove the `application` key, as `False` is the default value.

huguesdk (approved):
simple and effective, with tests and all. 👍

resource_booking_simultaneous_human/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+When a resource of type human is scheduled for a booking, they remain
+simultaneously available for other bookings.
```
pedrobaeza, line 2: Is this sentence a fact without this module, or the goal of this one?

carmenbianca: Added the word 'will'.

resource_booking_simultaneous_human/__manifest__.py
carmenbianca: Done

ntsirintanis (approved):

PR #79, opened 2023-02-03
[14.0][IMP] add tests assigning multiple human resources

huguesdk:
~~ensure multiple human resources can be assigned to a booking. the code handling the auto-confirmation of the calendar event assumed that there was only one human resource assigned.~~

edit: the fix has been done in #88. this pr now only adds the tests.

### DIFF
--- a/resource_booking/tests/test_backend.py
+++ b/resource_booking/tests/test_backend.py
@@ -888,3 +888,82 @@ class BackendCase(SavepointCase):
                 utc.localize(datetime(2021, 3, 14, 2, 0)),
             )
         )
+
+    def test_assign_multiple_humans(self):
+        """
+        It should be possible to assign multiple humans to a booking.
+        """
+        user_4 = self.env["res.users"].create(
+            {
+                "email": "user_4@example.com",
+                "login": "user_4",
+                "name": "User 4",
+            }
+        )
+        r_user_4 = self.env["resource.resource"].create(
+            {
+                "calendar_id": self.r_calendars[0].id,
+                "name": user_4.name,
+                "resource_type": "user",
+                "tz": "UTC",
+                "user_id": user_4.id,
+            }
+        )
+        rbc = self.env["resource.booking.combination"].create(
+            {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As this is modifying a module already landed, you should keep the compatibility with the old context key, although you add the new one.

I'm wondering if this is better to expand it changing the requester (`partner_id`) to a many2many to be able to expand in the core such option. What do you think?

huguesdk:
only a variable name local to a function has changed. i don’t think the context nor the backwards compatibility is affected, but maybe i’m missing something?

to be honest, i didn’t dive into this whole feature of autoconfirmation of resource attendees, i only fixed what seemed to me to be a bug while using the module in a normal way (reproduced by the tests). again, i might be missing something. i don’t currently have an opinion about changing that relationship to a many2many.

resource_booking/tests/test_backend.py
```diff
@@ -675,3 +675,73 @@ def test_event_show_as_free(self):
...
+        )
+        rbc = self.env["resource.booking.combination"].create(
+            {"resource_ids": [(6, 0, [self.r_users[0].id, r_user_3.id])]}
+        )
+        self.env["resource.booking"].create(
+            {
+                "partner_id": self.partner.id,
+                "start": "2021-03-01 08:00:00",
+                "type_id": self.rbt.id,
+                "combination_id": rbc.id,
```
carmenbianca, line 923: I'm a little confused by this. There's no `resource.booking.type.combination.rel` linking these two. Ought that not ring some alarm bells somewhere?

carmenbianca (approved):
LGTM, awesome tests @huguesdk

resource_booking/tests/test_backend.py
huguesdk: thanks for catching this! fixed.

houssine78 (approved):
LGTM

huguesdk:
since the changes to the code were made in #88, this pr now only adds the tests.

pedrobaeza:
/ocabot merge patch

Please fw-port these changes.

PR #81, opened 2023-02-10
[14.0][ADD] new module resource_booking_number_of_attendees

huguesdk:
this module only adds an integer field to resource bookings to store the number of attendees (independently of odoo meeting attendees). it can be useful for the organizer of bookings that involves groups of people to know the amount of (unmanaged) material should be prepared.

### DIFF
--- a/resource_booking_number_of_attendees/models/resource_booking.py
+++ b/resource_booking_number_of_attendees/models/resource_booking.py
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ResourceBooking(models.Model):
+    _inherit = "resource.booking"
+
+    # fixme: store this in the meeting instead of in the booking (see roadmap)
+    number_of_attendees = fields.Integer("Number of Attendees", default=1)
--- a/resource_booking_number_of_attendees/__init__.py
+++ b/resource_booking_number_of_attendees/__init__.py
@@ -1,0 +1,5 @@
+# SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from . import models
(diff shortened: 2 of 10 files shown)


carmenbianca (approved):
Everything here looks OK to me.

Does it maybe make sense to add the field on the meeting as a related field? Only show it when the meeting has a booking linked to it, and make sure the string + help text are sufficiently descriptive.

resource_booking_number_of_attendees/models/resource_booking.py
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ResourceBooking(models.Model):
+    _inherit = "resource.booking"
+
+    # fixme: store this in the meeting instead of in the booking (see roadmap)
+    number_of_attendees = fields.Integer("Number of Attendees", default=1)
```
pedrobaeza, line 12: A less typo-prone variable name is:

```suggestion
    attendee_number = fields.Integer("Number of Attendees", default=1)
```

pedrobaeza:
Why not including all the attendees in the booking instead (res.partner)?

carmenbianca:
@pedrobaeza I think the partners may not exist. If Jane Doe makes a booking for a training and says that she will show up with 4 people, then 3 of those people are not known to the people providing the training.

pedrobaeza:
Yeah, that's possible. Then this can be part also of the main module, and having this field as computed writable, being automatically incremented for the combination + attendees (if switched to m2m instead of m2o).

huguesdk:
i’m not sure i understand exactly what you mean. do you mean to:

1.  add this field on `calendar.event` (and a related field on the booking)
2.  compute it from both:
    1.  the number of attendees of the event (`partner_ids`)
    2.  the number of human resources of the combination
3.  have this field writable so the value can be set manually as well

which m2o relationship would you change to m2m? `resource.booking.partner_id`?

what to do in case the attendees change? should the possibly manual value be dropped and replaced by the computed value, or should a difference be computed?

do you think it is useful to take the human resources of the combination into account?

PR #82, opened 2023-02-10
[14.0][ADD] new module resource_booking_description

huguesdk:
this module only adds a text field to resource bookings to allow the organizer to add more information.


pedrobaeza:
I think this one can be part of the main module.

carmenbianca:
agreed with pedro

huguesdk:
change to main module done in #86.

PR #80, opened 2023-02-09
[14.0][IMP] resource_booking: Create resource.is_available() convenience function

carmenbianca:
This work will be a dependency for a module being created by @huguesdk.

Internal task: https://gestion.coopiteasy.be/web#id=9911&action=475&active_id=434&model=project.task&view_type=form&menu_id=536

PR #83, opened 2023-02-10
[14.0][ADD] add new module resource_booking_auto_create

huguesdk:
and also add new modules resource_category and resource_category_booking.

this module provides a wizard that allows to select available resources by category to assign to a booking. resource combinations are handled automatically.

this module depends on #80.


pedrobaeza:
Isn't better to add the category concept directly to the resource.booking model (so you can assign combinations or categories), and then adapt the selection code to take that into account?

carmenbianca:
@pedrobaeza I think it was written this way to stay compatible with the way of working in `resource_booking`. If the whole thing could be re-engineered from scratch (or if breaking compatibility is not a problem), it would probably be easier to remove the concept of 'combinations' `resource_booking`, and instead simply directly store the `resource.resource`s in `resource.booking`.

Subsequently, you could create modules that introduce the concept of combinations or categories to `resource.booking`.

Ripping out categories from `resource.booking` and replacing them with categories without doing the above refactoring seems like a recipe for bugs to me.

pedrobaeza:
What I'm thinking is complementary: just a new selector at resource.booking level to select also categories, and take into account all the resources of the combinations or the categories. They are minimal changes.

But as you prefer...

resource_booking_combination_auto_create/models/resource_booking.py
```diff
@@ -0,0 +1,36 @@
...
+    _inherit = "resource.booking"
+    # override constraint to allow to save resource bookings without
+    # combinations in pending state.
+    _sql_constraints = [
+        (
+            "combination_required_if_event",
+            "check(meeting_id is null or state = 'pending' or combination_id is not null)",
+            "Missing resource booking combination.",
+        ),
+    ]
```
carmenbianca: This should be in upstream `resource_booking` probably.

resource_booking_combination_auto_create/models/resource_booking.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _check_scheduling(self):
+        has_meeting = self.filtered("meeting_id")
+        if not has_meeting:
+            return
+        has_rbc = self.filtered("combination_id.resource_ids")
+        ready_rbc = has_meeting & has_rbc
+        # only check scheduling for bookings that have a meeting and a
+        # combination.
+        return super(ResourceBooking, ready_rbc)._check_scheduling()
```
carmenbianca, line 35: idem

resource_booking_combination_auto_create/models/resource_booking.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _compute_state(self):
+        for rb in self:
+            super(ResourceBooking, rb)._compute_state()
+            if rb.state == "scheduled" and not rb.combination_id:
+                # stay in pending state when a start date is set but no
+                # combination.
+                rb.state = "pending"
```
carmenbianca, line 43: idem

resource_booking_combination_auto_create/tests/test_resource_combination_wizard.py
```diff
@@ -0,0 +1,188 @@
...
+        already selected.
+        """
+
+    def test_no_categories_selected(self):
+        """
+        When no categories are selected, it should go to a special step that
+        displays an error message, and allows only to go back or cancel.
+        """
+
+    @freeze_time("2023-01-31 12:00:00")
```
carmenbianca: I think you can use the freeze_time decorator on the class instead, to save yourself some effort.

resource_booking_combination_auto_create/wizards/resource_booking_category_selection.py
```diff
@@ -0,0 +1,29 @@
...
+        required=True,
+        ondelete="cascade",
+    )
+    resource_category_id = fields.Many2one(
+        "resource.category",
+        "Resource Category",
+        required=True,
+        ondelete="cascade",
+    )
+    quantity = fields.Integer("Quantity", required=True, default=1)
```
carmenbianca: This needs a comment or a help field.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+    resource_booking_category_selection_ids = fields.One2many(
+        "resource.booking.category.selection",
+        "resource_booking_combination_wizard_id",
+        "Resource Categories",
+    )
+    current_resource_booking_category_selection_id = fields.Many2one(
+        "resource.booking.category.selection",
+        "Current Resource Category Selection",
+        ondelete="cascade",
+    )
```
carmenbianca: A comment to describe what 'current' means is welcome here.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+    current_resource_booking_category_selection_id_id = fields.Integer(
+        "Current Resource Category Selection ID",
+        related="current_resource_booking_category_selection_id.id",
+    )
```
carmenbianca: A comment to describe the necessity of this field would be nice.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+    current_resource_category_name = fields.Char(
+        related="current_resource_booking_category_selection_id"
+        ".resource_category_id.name",
+    )
```
carmenbianca: idem

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+        string="Resources",
+        related="current_resource_booking_category_selection_id.resource_ids",
+        readonly=False,
+    )
+    configure_step = fields.Integer("Configuration Step", default=0)
+    configure_step_count = fields.Integer(
+        "Number of Configuration Steps",
+        compute="_compute_configure_step_count",
+        store=True,
+    )
```
carmenbianca: I think I would prefer 'total' or 'amount' over 'count' here. I will leave this to you.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+    selected_resource_ids = fields.Many2many(
+        "resource.resource",
+        string="Selected Resources",
+        compute="_compute_selected_resource_ids",
+    )
+
+    @api.depends("resource_booking_category_selection_ids")
+    def _compute_configure_step_count(self):
+        self.configure_step_count = len(self.resource_booking_category_selection_ids)
```
carmenbianca: Isn't there a for-loop over `self` missing here?

Idem for all other compute methods.

Is this because wizards don't generally get grouped together in a recordset?

resource_booking_combination_auto_create/wizards/resource_booking_category_selection.py
```diff
@@ -0,0 +1,29 @@
...
+    resource_ids = fields.Many2many(
+        "resource.booking.category.selection.resource",
+        string="Resources",
+    )
```
carmenbianca: To avoid confusion, this should be renamed to `selection_resource_ids`.

I found this rather confusing in the following bit of code:

```python
    @api.depends("resource_booking_category_selection_ids.resource_ids.resource_id")
    def _compute_selected_resource_ids(self):
        self.selected_resource_ids = self.mapped(
            "resource_booking_category_selection_ids.resource_ids.resource_id"
        )
```

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+                    current_step=self.configure_step + 1,
+                    step_count=self.configure_step_count,
+                )
+            )
+        return _("Confirm")
+
+    def _reopen_self(self):
+        act_window = super()._reopen_self()
+        act_window["name"] = self._compute_step_name()
+        return act_window
```
carmenbianca: A comment describing what bug this fixes. I assume it's a workaround for something.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def _update_configure_step(self):
+        self.current_resource_booking_category_selection_id = (
+            self.resource_booking_category_selection_ids[self.configure_step]
+        )
```
carmenbianca: Is there a reason this isn't computed?

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+            self.state = "start"
+        else:
+            self.state = "configure"
+
+    def state_exit_configure(self):
+        if self.configure_step == self.configure_step_count - 1:
+            self.state = "final"
+        else:
+            self.configure_step += 1
+            self._update_configure_step()
```
carmenbianca: Comments :) I think what I'm missing is, for each method, why/when the method might be called. The subsequent code seems fairly self-evident.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def resource_is_available(self, resource_id):
+        # todo: implement this
+        return True
+
+    def find_available_resources(self):
```
carmenbianca: Method name lies. It finds-and-links available resources. I propose to either change the method name (something like 'update'), or to split into two methods if possible.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+                if not self.resource_is_available(resource_id):
+                    continue
+                rb_category_selection_resource_model.create(
+                    {
+                        "resource_booking_category_selection_id": rbcs.id,
+                        "resource_id": resource_id.id,
+                    }
+                )
+
+    def create_combination(self):
```
carmenbianca, line 289: Making this as a note without yet having read the contents: I'm very surprised that this takes no arguments.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -0,0 +1,209 @@
...
+        domain = []
+        for resource in resources:
+            domain.append(("resource_ids", "=", resource.id))
+        resource_combination_model = self.env["resource.booking.combination"]
+        existing_combinations = resource_combination_model.search(domain)
+        resource_combination = resource_combination_model.browse()
+        for resource_combination in existing_combinations:
+            # use first matching combination
+            if len(resource_combination.resource_ids) == len(resources):
+                break
```
carmenbianca: Does this do anything?

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -182,9 +182,10 @@ def create_combination(self):
         resource_combination_model = self.env["resource.booking.combination"]
         existing_combinations = resource_combination_model.search(domain)
         resource_combination = resource_combination_model.browse()
-        for resource_combination in existing_combinations:
+        for combination in existing_combinations:
             # use first matching combination
-            if len(resource_combination.resource_ids) == len(resources):
+            if len(combination.resource_ids) == len(resources):
+                resource_combination = combination
                 break
```
carmenbianca: Can't you use `(combination.resource_ids & resources) == resources`?

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
```diff
@@ -187,11 +187,16 @@ def find_available_resources(self):
 
     def create_combination(self):
         # search for an existing combination that matched the selected resources.
+        resource_combination_model = self.env["resource.booking.combination"]
         resources = self.selected_resource_ids
+        if not resources:
+            self.resource_booking_id.combination_id = (
+                resource_combination_model.browse()
+            )
+            return
```
carmenbianca: Why not set to False?

resource_booking_combination_auto_create/tests/test_resource_combination_wizard.py
```diff
@@ -301,6 +302,69 @@ def test_select_resource_in_multiple_selected_categories(self):
...
+        )
+        with self.assertRaises(ValidationError) as cm:
+            self._create_wizard_with_selected_categories(
+                booking, [self.room_category, self.worker_category]
+            )
+        self.assertEqual(
+            str(cm.exception),
+            "To select resources, the booking must have a start date and a "
+            "duration.",
+        )
```
carmenbianca, line 473: Did cm survive the context? Shouldn't this be indented to the right?

carmenbianca (changes requested):
Went through this commit-by-commit. It was a lot to go through; my level of scrutiny decreased during later commits.

Some comments may be made obsolete by later commits.

resource_booking_combination_auto_create/models/resource_booking.py
huguesdk: @pedrobaeza what do you think? this change is to allow to save a booking with a date but no resource combination, and leave it in `pending` state. currently this is not allowed by `resource_booking`, because as soon as a date is set, it considers the booking to be in `scheduled` state, and that requires a resource combination.

resource_booking_combination_auto_create/tests/test_resource_combination_wizard.py
huguesdk: done in a later commit.

resource_booking_combination_auto_create/wizards/resource_booking_category_selection.py
huguesdk: at first, the user would have to select a quantity of resources for each category, but this has been removed in a later commit.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
huguesdk: this field has been removed in a later commit. it was used in the wizard to display the resource category of the current configuration step, but this is now handled by the `current_configuration_step_message` field.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
huguesdk: you’re right. i’ll fix all compute methods.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
huguesdk: i didn’t like that name either. renamed to `num_configuration_steps`. what do you think?

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
huguesdk: the domain already ensures that the combination contains all the needed resources, but it can contain more. this is why comparing the length is enough and slightly more efficient.

resource_booking_combination_auto_create/tests/test_resource_combination_wizard.py
huguesdk: this seems very weird indeed, but that code would not be reachable if it would be indented to the right, as the context is exited as soon as the exception is raised. do you see another way of checking the message of the exception?

huguesdk:
thanks a lot @carmenbianca for this in-depth review. all the changes you suggested are now made. i also renamed the wizard models to shorter and more understandable names.

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
carmenbianca: Perfect!

resource_booking_combination_auto_create/wizards/resource_booking_combination_wizard.py
carmenbianca: ha, that's really cool

resource_booking_combination_auto_create/tests/test_resource_combination_wizard.py
carmenbianca: from a little searching on the internet; apparently this is completely normal. see also other example.

```python
🐚 coopiteasy reuse-tool main $ with open("README.md") as fp:
...............................     text = fp.read()
...............................
🐚 coopiteasy reuse-tool main $ fp
<_io.TextIOWrapper name='README.md' mode='r' encoding='UTF-8'>
🐚 coopiteasy reuse-tool main $ fp.closed
True
```

houssine78 (approved):
LGTM

PR #84, opened 2023-02-13
[14.0][ADD] add new module calendar_resource_booking_type_color

huguesdk:


pedrobaeza:
I think this can be an improvement of the main module instead of going with a new module.

huguesdk:
in the main module, the color of calendar events depend on the resource booking combination. are you suggesting to change this to use the booking type instead, or having the option to choose?

pedrobaeza:
Where are you talking about that? On the calendar view of the resource bookings? I think such colors are not even in the legend, isn't it?

huguesdk:
yes, it is about the color of events on the calendar view. currently they depend on `combination_id`: see [resource_booking/views/resource_booking_views.xml:16](https://github.com/OCA/calendar/blob/14.0/resource_booking/views/resource_booking_views.xml#L16).

pedrobaeza:
Yeah, but then no legend on the left part about what that color means... I think we can improve this in general. Putting avatars for the participants and color for types seems good.

carmenbianca (approved):

huguesdk:
change to main module done in #96.

PR #85, opened 2023-02-14
[ADD] resource_booking_remove_default_filter

carmenbianca:
Replaces https://github.com/coopiteasy/addons/pull/278

Internal task: https://gestion.coopiteasy.be/web#id=9873&action=475&active_id=434&model=project.task&view_type=form&menu_id=536


huguesdk (approved):

pedrobaeza:
I think a module for this feature seems too much. In the DB, you can:

- Remove the filter, and save it as a default favorite called for example "All".
- Go to user filters in configuration, and remove the user.

That way, you get the same result without having an explicit module for that.

Another option may be to have 2 menu entries, one with the filter, called "My bookings", and another general without it, and defaulting to this second one.

PR #86, opened 2023-02-17
[14.0][IMP] add description field to bookings

huguesdk:
add a description text field to resource bookings to allow the organizer to add more information.

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -63,6 +63,7 @@ class ResourceBooking(models.Model):
         ),
     )
     name = fields.Char(index=True, help="Leave empty to autogenerate a booking name.")
+    description = fields.Html("Description")
     partner_id = fields.Many2one(
         "res.partner",
         string="Requester",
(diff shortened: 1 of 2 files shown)


resource_booking/models/resource_booking.py
```diff
@@ -63,6 +63,7 @@ class ResourceBooking(models.Model):
         ),
     )
     name = fields.Char(index=True, help="Leave empty to autogenerate a booking name.")
+    description = fields.Text("Description")
```
pedrobaeza: Why not HTML to have rich text format?

cathLemb: The client didn't specify, so it's ok for me to change to html.

huguesdk: done.

pedrobaeza (approved):

carmenbianca (approved):
might be good practice to add a changelog. otherwise it's a good simple change :) lgtm

pedrobaeza:
I don't like changelog, as it's double work, and inconsistencies between both "source of truth". I prefer very good commit messages instead.

/ocabot merge minor

huguesdk:
thanks, @pedrobaeza!

PR #91, opened 2023-03-06
[14.0][FIX] resource_booking: Force to send calendar event author notification.

victoralmau:
Force to send calendar event author notification (https://github.com/odoo/odoo/blob/14.0/addons/mail/models/mail_thread.py#L2470).

Steps to reproduce:

- Create a booking
- Set a manual assigned
- Share + confirm (with portal)
- User has been notified

@Tecnativa TT40302

### DIFF
--- a/resource_booking/models/calendar_event.py
+++ b/resource_booking/models/calendar_event.py
@@ -57,6 +57,8 @@ class CalendarEvent(models.Model):
         """Transfer resource booking to _attendees_values by context.
 
         We need to serialize the creation in that case.
+        mail_notify_author key from context is necessary to force the notification
+        to be sent to author.
         """
         vals_list2 = []
         records = self.env["calendar.event"]
@@ -65,7 +67,8 @@ class CalendarEvent(models.Model):
                 records += super(
                     CalendarEvent,
                     self.with_context(
-                        resource_booking_ids=vals["resource_booking_ids"]
+                        resource_booking_ids=vals["resource_booking_ids"],
+                        mail_notify_author=True,
                     ),
                 ).create(vals)
             else:


resource_booking/models/resource_booking.py
```diff
@@ -532,29 +532,43 @@ def name_get(self):
             new.append((id_, name))
         return new
 
+    def _get_combination_from_updated_values(self, updated_values):
```
pedrobaeza: Why extracting a method? Is it used in another part?

resource_booking/models/resource_booking.py
```diff
@@ -532,29 +532,43 @@ def name_get(self):
...
-                    partner.id,
-                    default_subtype_ids,
-                    "resource_booking.message_combination_assigned"
-                    if partner != self.env.user.partner_id
-                    else False,
+        combination = self._get_combination_from_updated_values(updated_values)
+        if combination:
+            resource_partners = combination.mapped("resource_ids.user_id.partner_id")
+            for partner in resource_partners.filtered("active"):
+                result.append(
```
pedrobaeza: Doing this they receive the booking details? What if they aren't subscribed by default to notes?

pedrobaeza (commented):
Please check in v13, that the notifications were done by calendar module standard calendar event mail template and mechanisms

victoralmau:
> Please check in v13, that the notifications were done by calendar module standard calendar event mail template and mechanisms

Sorry, but if I try v13 I don't get an email in the booking confirmation either.

pedrobaeza:
I was getting it from calendar.event. Probably the problem is that you are confirming the booking from your own user. Try to book it from the portal in an incognito window.

victoralmau:
Ok, it has been checked and the necessary fix is now applied.

pedrobaeza (approved):
Tested in a real environment, and working as expected. Fast-tracking this to solve the problem:

/ocabot merge patch

PR #98, opened 2023-07-17
[13.0][FIX] resource_booking: remove test dependency from manifest

LoisRForgeFlow:
I think test dependencies should not be included in the manifest, otherwise it is forcing to install them in production environments where it is not really needed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
 # generated from manifests external_dependencies
 cssselect
-freezegun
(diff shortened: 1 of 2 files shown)


yajo:
I think the opposite. We don't have a place where to record non-production dependencies. So if there's no other place, better one than none.

LoisRForgeFlow:
I guess it is a matter of taste and your opinion is more important as original author. Thanks for reviewing, great module BTW!

pedrobaeza:
The ideal is to have an extra manifest key `external_dependencies_tests` or similar for this, and be parsed by OCA tools to put `test_requirements.txt`. The good news is that this specific library is already included in Odoo core in upper versions, and no need to put it here.

yajo:
That'd be nice, yes.

PR #104, opened 2023-09-26
[15.0][FIX] resource_booking: Add class to schedule page to show full dropdown if opened at the bottom

victoralmau:
Add class to schedule page to show full dropdown if opened at the bottom

**Before**
antes

**After**
despues

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT45274

### DIFF
--- a/resource_booking/templates/portal.xml
+++ b/resource_booking/templates/portal.xml
@@ -545,6 +545,7 @@
             </t>
             <!-- Scheduling form -->
             <t t-call="portal.portal_record_layout">
+                <t t-set="classes" t-value="'resource_booking_portal_schedule'" />
                 <t t-set="card_header">
                     <t t-call="resource_booking.resource_booking_portal_header" />
                 </t>
--- a/resource_booking/static/src/scss/portal.scss
+++ b/resource_booking/static/src/scss/portal.scss
@@ -6,3 +6,7 @@
     max-height: 40vh;
     overflow: auto;
 }
+// Show full dropdown if opened at the bottom
+.resource_booking_portal_schedule {
+    overflow: initial;
+}


resource_booking/templates/portal.xml
```diff
@@ -545,6 +545,7 @@
             </t>
             <!-- Scheduling form -->
             <t t-call="portal.portal_record_layout">
+                <t t-set="classes" t-value="'resource_booking_portal_schedule'" />
```
pedrobaeza, line 548: Isn't this overwriting other existing (or inherited) classes?

pedrobaeza:
@norlinhenrik I think you had this problem too in v16.

resource_booking/templates/portal.xml
victoralmau: This only adds extra classes: https://github.com/odoo/odoo/blob/15.0/addons/portal/views/portal_templates.xml#L323

pedrobaeza: Yes, but what happens if other module set the same variable? Yours will be overwritten, or vice versa.

pilarvargas-tecnativa: The problem is that there is no identifying selector to apply the css, the only alternative solution would be to inherit the template, xpath to the element with the .card class and add the new class by modifying the class attribute. This way if the template is called from somewhere else the class will not disappear. I have tried to play with the .slots-dropdown class which is quite identifiable and corresponds to the dropdown, the problem is that this class is conditioned by the .card class of the parent container and as there is nothing identifiable to generate a selector to this class (.card) which is quite general, it cannot be modified freely. Another option would be to define a small script in js to select the container element of the .slots-dropdown class but I think that is quite radical and I don't know if in this case it would be more optimal to inherit the template.

pedrobaeza: If the problem is to have a specific selector, the view is built from this module, so we can put any other needed class/id.

pilarvargas-tecnativa: I've been testing, in particular I've set a red background class in another call to the same template and each call maintains its own custom classes. I think that what @victoralmau proposes is quite coherent having the option to modify the class. In this way the class is applied at the call level, so if from another place they make a call to the same template and modify the classes it won't affect this particular call but it won't propagate to other calls either. That is to say, what is established in the t-set will have repercussions on this specific call, it can neither be modified from outside and cannot be propagated to other calls.

pedrobaeza: OK then.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #105, opened 2023-09-27
[ADD] calendar_attachment: Show attachments in modal window

tarteo:
The calendar view opens the form in a modal window therefore the chatter is not visible.
This module adds a tab with the attachments on the event form, so the attachments are
also accessible from the calendar view.

image


pedrobaeza:
Isn't better as a general approach to include a way to open the event full form from the popup or the dropdown?

tarteo:
Hey @pedrobaeza, If it opens a new window or uses the current window depends on how it's implemented e.g. sale.order calendar opens the record in the current window. 
Even then, the customer wants to use the modal window for the calendar app (calendar.event).

pedrobaeza:
OK, it was just a suggestion specific to `calendar.event`. You can go this way.

ICTRV:
this looks good so far

PR #110, opened 2023-11-07
[15.0][IMP] resource_booking: Add multiples requesters in booking

carolinafernandez-tecnativa:
- Add multiples requesters in booking

@Tecnativa
TT45338

@pedrobaeza


resource_booking/models/resource_booking.py
```diff
@@ -125,7 +125,12 @@ class ResourceBooking(models.Model):
     description = fields.Html()
     partner_id = fields.Many2one(
         comodel_name="res.partner",
+        compute="_compute_partner_id",
         string="Requester",
+    )
+    partner_ids = fields.Many2many(
+        comodel_name="res.partner",
+        string="Requesters",
         index=True,
```
pedrobaeza: `index` doesn't make sense in m2m, nor `ondelete`.

resource_booking/models/resource_booking.py
```diff
@@ -674,11 +686,11 @@ def action_confirm(self):
                 if not booking.meeting_id:
                     continue
                 # Make sure requester and user resources are meeting attendees
```
pedrobaeza:
```suggestion
                # Make sure requesters and user resources are meeting attendees
```

resource_booking/models/resource_booking.py
carolinafernandez-tecnativa: remove index=True. Should i keep ondelete?

pedrobaeza: Please remove it

pedrobaeza:
This is not ADD, but IMP, and the commit message shouldn't contain the version.

carolinafernandez-tecnativa:
> This is not ADD, but IMP, and the commit message shouldn't contain the version.
Done, thanks!

migrations/15.0.1.4.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2023 Tecnativa - Carolina Fernandez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from psycopg2.extensions import AsIs
+
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    """Put partner_id in partner_ids"""
+    env = api.Environment(cr, SUPERUSER_ID, {})
```
pedrobaeza: Use `openupgradelib.m2o_to_x2x` method instead.

carolinafernandez-tecnativa: could you please check it now? Thanks!

migrations/15.0.1.4.0/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Tecnativa - Carolina Fernandez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(env):
+    env.cr.execute(
+        """
+        ALTER TABLE resource_booking RENAME partner_id TO migrated_partner_id
```
pedrobaeza: Better to use a familiar nomenclature (and more expressive):

```suggestion
        ALTER TABLE resource_booking RENAME partner_id TO old_partner_id
```

pedrobaeza (commented):
I'm thinking that now that we put several partners, we should switch the nomenclature from "Requester" to "Attendees". What do you think?

migrations/15.0.1.4.0/pre-migration.py
pedrobaeza: And technically FYI, there's no need to rename the column, as this column is available on post-migration.

carolinafernandez-tecnativa: Ok,  so if the column is available i deleted the file pre-migration.py

carolinafernandez-tecnativa:
> I'm thinking that now that we put several partners, we should switch the nomenclature from "Requester" to "Attendees". What do you think?

I agree. I just changed :)

migrations/15.0.1.4.0/pre-migration.py
pedrobaeza: Oh, wait, I didn't see that you are still defining `partner_id` as a regular field, but not stored, so in this case the column is removed, so please rename. Sorry for not seeing before that issue.

pedrobaeza (approved):
Thanks for the new feature:

/ocabot merge nobump

pedrobaeza:
Please review 16.0 migration and then let's include this one after the merge.

PR #118, opened 2024-02-22
[15.0][ADD] calendar_import_ics

ArnauCForgeFlow:
This module adds a new wizard that allows you to import .ics files into odoo calendar.


calendar_import_ics/models/calendar_event.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (C) 2024 - ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class CalendarEvent(models.Model):
+    _inherit = "calendar.event"
+    import_id = fields.Char("Imported Event Id")
```
LoisRForgeFlow: Can you rename this to `import_identifier` or `event_identifier`? _id is usually reserverd to m2o fields (by convention).

in the label you can use "Event ID" or "Event Identifier"

LoisRForgeFlow (approved):
Functional review LGTM

PR #101, opened 2023-08-30
[16.0][MIG] resource_booking: Migration to 16.0

norlinhenrik:
Improvements in this migration:

- Booking: Refactored _get_available_slots() works also when each slot is multiple days
- Booking Type: new field slot_duration
- Contact: button -> Bookings
- Combination -> Bookings -> Create booking: Set the default combination from the navigation
- Booking: may have multiple contacts
- Booking: When type_id is missing (booking from Timeline view), get empty Intervals (instead of error)
- Booking: Search on combination
- Booking: Refactored _availability_is_fitting()
- Booking: List view with hidden contacts (the new field)
- Work intervals (with migration): To span multiple days, set hour_to = 24 instead of 23:59
- XML renaming (with migration)
- #100 

Later PRs for version 15.0 of `resource_booking` (104, 106, 110, 111) are not included in this PR.

PR #120, opened 2024-03-11
[16.0][MIG] resource_booking: Migration to 16.0

victoralmau:
Superseed https://github.com/OCA/calendar/pull/101

Migration to 16.0

Changes done:
- [x] Squash administrative commits.
- [x] Purge translations.
- [x] Fix tests and/or remove warnings.

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa TT45643


resource_booking/tests/test_portal.py
```diff
@@ -7,33 +7,29 @@
 from freezegun import freeze_time
 from lxml.html import fromstring
 
+import odoo
```
pedrobaeza: Avoid this general import importing and using just `tagged`

resource_booking/tests/test_backend.py
```diff
@@ -645,22 +645,22 @@ def test_videocall_location(self):
         with Form(rb) as rb_f:
             rb_f.start = "2021-03-01 08:00:00"
         self.assertEqual(rb.state, "scheduled")
-        self.assertEqual(rb.videocall_location, "Videocall Office 1")
-        self.assertEqual(rb.meeting_id.videocall_location, "Videocall Office 1")
+        self.assertTrue("Videocall Office 1" in rb.videocall_location)
```
pedrobaeza: Use `assertIn` instead (in all occurrences). Anyway, I don't get why you are changing this.

resource_booking/tests/test_backend.py
```diff
@@ -268,9 +268,9 @@ def test_booking_from_calendar_view(self):
         self.assertEqual(self.rbt.duration, 0.5)
         # Change it to 45 minutes
         self.rbt.duration = 0.75
-        # Bookings smart button configures calendar with slots of 45 minutes
+        # Bookings smart button configures calendar with slots from slot duration field
         button_context = self.rbt.action_open_bookings()["context"]
-        self.assertEqual(button_context["calendar_slot_duration"], "00:45")
+        self.assertEqual(button_context["calendar_slot_duration"], "00:30:00")
```
pedrobaeza, line 273: Why this change in the test?

resource_booking/migrations/16.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
+from openupgradelib import openupgrade
+
+
+def _update_attendance_hour_to(env):
+    # 23:59 -> 24:00
+    lines = env["resource.calendar.attendance"].search(
```
pedrobaeza, line 10: Why is this needed?

resource_booking/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,33 @@
+from openupgradelib import openupgrade
+
+xmlids_spec = [
```
pedrobaeza: Why renaming this?

resource_booking/models/resource_booking.py
```diff
@@ -137,6 +172,14 @@ class ResourceBooking(models.Model):
         tracking=True,
         help="Who will attend this booking?",
     )
+    partner_ids = fields.Many2many(
```
pedrobaeza: This one shouldn't be added in this commit.

resource_booking/tests/common.py
```diff
@@ -42,7 +42,7 @@ def create_test_data(obj):
                 "name": "Fridays",
                 "dayofweek": "4",
                 "hour_from": 0,
-                "hour_to": 23.99,
+                "hour_to": 24,
```
pedrobaeza, line 45: This should go in the same direction as the migration script, but I don't understand why it's needed.

pedrobaeza (changes requested):
I think you can merge both commits of Henrik, and put a better first line commit message on the other.

resource_booking/tests/test_backend.py
victoralmau: I have left `assertEqual`, although it is now necessary to specify http or https to avoid auto-defining with the base url.

resource_booking/migrations/16.0.1.0.0/pre-migration.py
victoralmau: Removed.

resource_booking/models/resource_booking.py
victoralmau: The final commit I added removes this (I didn't want to modify the author's commits).

pedrobaeza: I think this is confusing in this case. Better to not change it. And this should be already existing in 15.0 previous commits.

resource_booking/tests/test_backend.py
victoralmau: All this change is related to the `slot_duration` field that is added in `resource.booking.type`, it changes with respect to how it is set in v15: https://github.com/OCA/calendar/blob/15.0/resource_booking/models/resource_booking_type.py#L179 Is it necessary? I think not because there are now 2 fields to reflect the same thing.

resource_booking/models/resource_booking.py
victoralmau: Ok, I will remove it from the corresponding commit

resource_booking/tests/test_backend.py
pedrobaeza: Well, that slot is though for presenting the calendar in intervals of such duration, so it has a goal, but not sure how it was handled before and now.

victoralmau: In v15 this already exists (with the `duration` field), so I comment that I don't understand the need for this extra field for "the same thing".

victoralmau: Ok, I see that it is related to https://github.com/OCA/calendar/pull/101#discussion_r1374273498 and is intentional.

resource_booking/tests/common.py
victoralmau: This change checks https://github.com/OCA/calendar/pull/120#discussion_r1519952482

resource_booking/migrations/16.0.1.0.0/post-migration.py
victoralmau: I have added a comment in the method to try to explain why this is necessary.

victoralmau:
Changes done.

jelenapoblet (approved):
Tested on runboat and on premise instance, working fine on my side.

carolinafernandez-tecnativa:
I have tested, and after installing contacts module i got this error message

image

It is refering to this field: https://github.com/OCA/calendar/pull/120/commits/b31f3136c0987800f7affeb8cbfcf15c89768730#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R10

https://github.com/OCA/calendar/pull/120/commits/b31f3136c0987800f7affeb8cbfcf15c89768730#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R18

On this PR, there was a change to allow multiple partners in resource booking.
https://github.com/OCA/calendar/pull/110

Thanks!

victoralmau:
> I have tested, and after installing contacts module i got this error message
> image
> 
> It is refering to this field: [b31f313#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R10](https://github.com/OCA/calendar/commit/b31f3136c0987800f7affeb8cbfcf15c89768730#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R10)
> 
> [b31f313#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R18](https://github.com/OCA/calendar/commit/b31f3136c0987800f7affeb8cbfcf15c89768730#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R18)
> 
> On this PR, there was a change to allow multiple partners in resource booking. #110
> 
> Thanks!

FIxed.

carolinafernandez-tecnativa:
> > I have tested, and after installing contacts module i got this error message
> > image
> > It is refering to this field: [b31f313#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R10](https://github.com/OCA/calendar/commit/b31f3136c0987800f7affeb8cbfcf15c89768730#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R10)
> > [b31f313#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R18](https://github.com/OCA/calendar/commit/b31f3136c0987800f7affeb8cbfcf15c89768730#diff-b082390b74a6351a8e33162e48a73b9ee0840757e7a7006440ca5eea629705c1R18)
> > On this PR, there was a change to allow multiple partners in resource booking. #110
> > Thanks!
> 
> FIxed.

Thanks, it is solved.

When im trying to book for this resource, i got this error message:

image


Check this configuration:

image

image

image

Maybe it there any conflict with timezone?

victoralmau:
Checking on this, I see that there is no problem, although there is some confusion.

In runboat the Admin user has the Timezone America/Montevideo, therefore, when you want to set a start date for the booking, you are doing it with that date and time in America/Montevideo, but if we "_convert_" the date and time it would not be available in the corresponding work schedule.

carolinafernandez-tecnativa (approved):
Functional Review, LGTM :) Thanks!

resource_booking/models/res_partner.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    resource_booking_count = fields.Integer(
+        compute="_compute_resource_booking_count", string="Resource booking count"
+    )
+    resource_booking_ids = fields.One2many(
+        "resource.booking", "partner_id", string="Bookings"
```
pedrobaeza: This one is incorrect, as it only reflects the first attendee. It should be a computed one.

pedrobaeza (commented):
The commit message `[IMP] resource_booking: add freezegun as a python external dependency` is incorrect, as it's the first one of a lot of changes. You should put a general commit message and then the summary. The same for the other one.

resource_booking/models/res_partner.py
victoralmau: Change to many2many field + default_partner_ids.

victoralmau:
> The commit message `[IMP] resource_booking: add freezegun as a python external dependency` is incorrect, as it's the first one of a lot of changes. You should put a general commit message and then the summary. The same for the other one.

Commit message changed (summary + changes).

norlinhenrik (approved):

pedrobaeza (approved):
/ocabot merge nobump

tarteo:
@victoralmau Why did you change the ecma version?

pedrobaeza:
Restored in #165

PR #121, opened 2024-05-17
[16.0][FIX] resource_booking: Add booking_count to portal values only if requested

victoralmau:
Add `booking_count` to portal values only if requested

The `_prepare_home_portal_values()` method sometimes requests `booking_count` and only then should we return the value.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa


resource_booking/templates/portal.xml
```diff
@@ -257,7 +257,7 @@
         </xpath>
     </template>
     <template
-        id="portal_my_home"
+        id="portal_my_home_resource_booking"
```
pedrobaeza: No need of changing this id. In fact, the usual rule is to keep the parent id (but within this module).

pedrobaeza (approved):
/ocabot merge patch

PR #124, opened 2024-05-21
[15.0][ADD] calendar_export_ics

ArnauCForgeFlow:
This module adds a new wizard that allows you to export odoo calendar to .ics file.


calendar_export_ics/wizards/wizard_export_ics.py
```diff
@@ -0,0 +1,64 @@
...
+        return {
+            "type": "ir.actions.act_window",
+            "view_mode": "form",
+            "res_model": "calendar.export.ics",
+            "name": "Export to Ics File",
+            "target": "new",
+            "res_id": self.id,
+        }
+
+    def generate_ics_content(self):
```
amh-mw, line 48: This function seems a bit DRY with respect to odoo/odoo `calendar.event` _get_ics_file
https://github.com/odoo/odoo/blob/f219583052e46ae8942675495c3833bd432dacd8/addons/calendar/models/calendar_event.py#L1056-L1114. Given that it does not make use of the `vobject` package, are there any concerns that this introduces .ics syntax errors or attack surfaces for injection?

amh-mw (changes requested):

calendar_export_ics/wizards/wizard_export_ics.py
amh-mw: See https://github.com/odoo/odoo/pull/166282 for an attempt to avoid DRY.

ArnauCForgeFlow: Hi @amh-mw,
You are right, it looks a bit DRY, i didn't know about the function _get_ics_file. 
Now the code uses this function to avoid DRY

amh-mw (approved):
LGTM WRT code review and unit test coverage. Have not tested in user interface.

LoisRForgeFlow (commented):
Tested, overall LGTM. One UX improvement that you could do it is to hide the generate button and configuration once the calendar is ready to download, it would make more easy to understand that it is ready to download

image

LoisRForgeFlow (approved):
/ocabot merge nobump

PR #127, opened 2024-05-24
[16.0][IMP] resource_booking: Add support for mail_note subtype by default

victoralmau:
Add support for mail_note subtype by default

If the default mail_note subtype is defined, when generating the calendar.event is generated and emails are sent to attendees, the mail_notify_author because each email would also be sent to the author (the event organizer).

Now it is changed to use a specific context and only send the email to the author if it is the the recipient of the email (the email of their attendance to the event).

Related to https://github.com/OCA/calendar/commit/575d2aa6d24d26ead89995d96e9709cdd5d9975a

@Tecnativa TT49045

### DIFF
--- a/resource_booking/models/calendar_event.py
+++ b/resource_booking/models/calendar_event.py
@@ -52,13 +52,32 @@ class CalendarEvent(models.Model):
         rescheduled._validate_booking_modifications()
         return result
 
+    def _notify_thread(self, message, msg_vals=False, **kwargs):
+        """If we are creating the calendar event from the resource booking
+        (detected from the resource_booking_event context key), we need to
+        inject the standard mail context `mail_notify_author` to super to get
+        the own author notified when someone books a reservation, but only
+        in the case that the mail is being sent to them, as if not the author
+        may receive one copy per each of the attendees. This happens only when
+        the the subtype not is enabled by default in the instance.
+        """
+        if self.env.context.get("resource_booking_event") and msg_vals.get(
+            "author_id"
+        ) in msg_vals.get("partner_ids", []):
+            self = self.with_context(mail_notify_author=True)
+        return super()._notify_thread(message=message, msg_vals=msg_vals, **kwargs)
+
     @api.model_create_multi
     def create(self, vals_list):
         """Transfer resource booking to _attendees_values by context.
 
         We need to serialize the creation in that case.
-        mail_notify_author key from context is necessary to force the notification
-        to be sent to author.
+        resource_booking_event custom key from context is necessary.
+        We cannot use mail_notify_author key in the context because if the mail_note
+        subtype is set by default, the email of each attendee would be sent also to
+        the author (example: a meeting with 2 attendees would send 2 emails but
+        each of them would be sent to the partner of the attendee + author of
+        the email).
         """
         vals_list2 = []
         records = self.env["calendar.event"]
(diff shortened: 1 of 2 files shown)


resource_booking/models/calendar_event.py
```diff
@@ -52,14 +52,28 @@ def write(self, vals):
...
 
+    def _notify_thread(self, message, msg_vals=False, **kwargs):
+        """We intercept this method because we can not overwrite the
+        _send_mail_to_attendees() method in calendar.attendee because it is done
+        with all the attendees and there we cannot indicate the mail_notify_author
+        context, if we did it, the email of each attendee would be sent additionally
+        to the author."""
+        if self.env.context.get("mail_notify_only_attendee_author"):
+            if msg_vals.get("author_id") in msg_vals.get("partner_ids"):
+                self = self.with_context(mail_notify_author=True)
```
pedrobaeza: But in this specific case we will have the same problem receiving all the emails, isn't it?

pedrobaeza:
The main comment in the PR doesn't correspond to the code.

victoralmau:
> The main comment in the PR doesn't correspond to the code.

I don't understand exactly what you mean, the code change and the description explain the specific use case (nothing like this happens if the mail_note subtype is not defined by default).

resource_booking/models/calendar_event.py
victoralmau: This problem **only** happens when creating a calendar event linked to `resource.booking` because we want it to notify all attendees (including the author of the email = organizer) but we do NOT want the organizer to receive all emails that are sent to all attendees.

resource_booking/models/calendar_event.py
```diff
@@ -68,7 +82,7 @@ def create(self, vals_list):
                     CalendarEvent,
                     self.with_context(
                         resource_booking_ids=vals["resource_booking_ids"],
-                        mail_notify_author=True,
+                        mail_notify_only_attendee_author=True,
```
pedrobaeza:
```suggestion
                        resource_booking_event=True,
```

resource_booking/models/calendar_event.py
```diff
@@ -52,14 +52,28 @@ def write(self, vals):
         rescheduled._validate_booking_modifications()
         return result
 
+    def _notify_thread(self, message, msg_vals=False, **kwargs):
+        """We intercept this method because we can not overwrite the
+        _send_mail_to_attendees() method in calendar.attendee because it is done
+        with all the attendees and there we cannot indicate the mail_notify_author
+        context, if we did it, the email of each attendee would be sent additionally
+        to the author."""
```
pedrobaeza:
```suggestion
        """If we are creating the calendar event from the resource booking
        (detected from the resource_booking_event context key), we need to
        inject the standard mail context `mail_notify_author` to super to get
        the own author notified when someone books a reservation, but only
        in the case that the mail is being sent to them, as if not the author
        may receive one copy per each of the attendees. This happens only when
        the the subtype not is enabled by default in the instance.
        """
```

resource_booking/models/calendar_event.py
```diff
@@ -52,14 +52,28 @@ def write(self, vals):
...
+        if self.env.context.get("mail_notify_only_attendee_author"):
+            if msg_vals.get("author_id") in msg_vals.get("partner_ids"):
```
pedrobaeza:
```suggestion
        if (
            self.env.context.get("resource_booking_event"):
            and msg_vals.get("author_id") in msg_vals.get("partner_ids", [])
        ):
```

victoralmau:
Changes done.

pedrobaeza (approved):
Seen directly that this is needed for avoiding duplicated mails in the case of notes are activated by default, so let's fast-track it:

/ocabot merge patch

PR #129, opened 2024-06-18
[16.0][IMP] resource_booking: Add support for mail_note subtype by default

victoralmau:
Add support for mail_note subtype by default

If we are creating the calendar event from resource booking, we want to notify only the partner_ids and not all followers (to avoid that each email is sent to all followers). Example: Resource booking with combination of several users. This happens only when the the subtype not is enabled by default in the instance.

Related to https://github.com/OCA/calendar/pull/127/commits/a3cd5a883b809237150e0da3247f68f02f4b4c4c

@Tecnativa TT49045

### DIFF
--- a/resource_booking/models/calendar_event.py
+++ b/resource_booking/models/calendar_event.py
@@ -67,6 +67,25 @@ class CalendarEvent(models.Model):
             self = self.with_context(mail_notify_author=True)
         return super()._notify_thread(message=message, msg_vals=msg_vals, **kwargs)
 
+    def _notify_get_recipients(self, message, msg_vals, **kwargs):
+        """If we are creating the calendar event from resource booking, we want to
+        notify only the partner_ids and not all the followers (to avoid that each email
+        is sent to all followers). Example: Resource booking with combination of several
+        users. This only happens when the subtype note is enabled by default in the
+        instance.
+        """
+        res = super()._notify_get_recipients(
+            message=message, msg_vals=msg_vals, **kwargs
+        )
+        if self.env.context.get("resource_booking_event"):
+            res2 = []
+            partner_ids = msg_vals.get("partner_ids", [])
+            for item in res:
+                if item["id"] in partner_ids:
+                    res2.append(item)
+            return res2
+        return res
+
     @api.model_create_multi
     def create(self, vals_list):
         """Transfer resource booking to _attendees_values by context.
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Isn't better to delay the subscription of the resources for avoiding both problems?

On the commit message, please change `related to` by `complement of`, as it's not the same.

victoralmau:
> Isn't better to delay the subscription of the resources for avoiding both problems?
> 
> On the commit message, please change `related to` by `complement of`, as it's not the same.

Followers are added when creating the `calendar.attendee` (https://github.com/odoo/odoo/blob/16.0/addons/calendar/models/calendar_attendee.py#L69C19-L69C37), if you prefer we can do that approach (delete the existing followers of mail_note subtype from all attendee created before and add them later).

pedrobaeza:
What I still don't understand is why doing a manual calendar event with multiple attendees, this is not happening. Why can't we mimic what standard does there?

victoralmau:
Yes, it is possible for this to happen with a manual event.

Steps to reproduce it.
- Activate the Notes subtype by default.
- Create a calendar event with future date and add to: Admin + Demo + Azure
- The invitation email from Azure will have been sent to Demo as well.

example

Do you think this is something that should be resolved in `calendar`?

resource_booking/models/calendar_event.py
```diff
@@ -67,6 +67,25 @@ def _notify_thread(self, message, msg_vals=False, **kwargs):
             self = self.with_context(mail_notify_author=True)
         return super()._notify_thread(message=message, msg_vals=msg_vals, **kwargs)
 
+    def _notify_get_recipients(self, message, msg_vals, **kwargs):
+        """If we are creating the calendar event from resource booking, we want to
+        notify only the partner_ids and not all followers (to avoid that each email is
```
pedrobaeza:
```suggestion
        notify only the partner_ids and not all the followers (to avoid that each email is
```

resource_booking/models/calendar_event.py
```diff
@@ -67,6 +67,25 @@ def _notify_thread(self, message, msg_vals=False, **kwargs):
...
+        sent to all followers). Example: Resource booking with combination of several
+        users. This happens only when the the subtype not is enabled by default in the
```
pedrobaeza:
```suggestion
        users. This only happens when the subtype note is enabled by default in the
```

pedrobaeza (approved):

pedrobaeza:
Let's keep it this way, as to get it upstream will be difficult, and mangle with followers will be difficult.

pedrobaeza (approved):
Let's see if we finally tackle this:

/ocabot merge patch

PR #130, opened 2024-06-26
[16.0][FIX] resource_booking: Change combination_id mandatory when autoassign is false

carolinafernandez-tecnativa:
- Change combination_id mandatory when autoassgin is false.
- Previously it was required when date is set but it is not taking into account if we share link and tried to book, user get an internal error server if combination_id is not set.

image

@Tecnativa
@pedrobaeza @victoralmau


pedrobaeza (approved):
Not sure about the CI error

victoralmau (approved):

PR #133, opened 2024-07-10
[16.0][IMP] resource_booking: Add search to partner_id field to prevent "wrong" results

victoralmau:
Add search to partner_id field to prevent "_wrong_" results

Please @pedrobaeza can you review it?

@Tecnativa TT50035

PR #134, opened 2024-07-10
[15.0][IMP] resource_booking: Add search to partner_id field to prevent "wrong" results

victoralmau:
Bacport from 16.0: https://github.com/OCA/calendar/pull/133

Add search to `partner_id` field to prevent "_wrong_" results + Add noupdate_changes.xml

@Tecnativa TT50035


pedrobaeza (approved):
Let's fix this data hole

/ocabot merge patch

PR #136, opened 2024-07-31
[16.0][IMP] resource_booking: get_intervals

norlinhenrik:
Current workflow: Create a booking with a booking type, then check availablility.

Needed workflow: Check availability for a booking type without creating a booking.

_get_intervals() has a new input argument:

- start_dt
- end_dt
- combination=None
- **type=None**

---

2nd commit: maintainers. Shall I squash the commits?


pedrobaeza (changes requested):
Don't use `type`, as it's a semi-reserved word. Use for example `rb_type`. Anyway, I don't get the sense of the code, as you are still assigning a booking type.

PR #141, opened 2024-12-16
[16.0][IMP] resource_booking: combination_id as filter on calendar view.

adasatorres:
Good morning, I would like to add this contribution to the resource_booking as I believe it is quite useful. The modification would allow filtering directly in the calendar view by the combination of resources, adding the `combination_id` field, similar to how the calendar addon works with contacts.

image

### DIFF
--- a/resource_booking/views/resource_booking_views.xml
+++ b/resource_booking/views/resource_booking_views.xml
@@ -18,7 +18,7 @@
                 <field name="name" invisible="True" />
                 <field name="partner_ids" widget="many2many_tags" />
                 <field name="type_id" />
-                <field name="combination_id" />
+                <field name="combination_id" filters="1" />
                 <field name="description" />
             </calendar>
         </field>


rrebollo (approved):
Code Review LGTM!

pedrobaeza (commented):
I don't have clear all the options of the field. Can you clarify them? I see also a lot of noise of adding the contributor just for this tiny change.

adasatorres:
Good morning, yesterday while reviewing in depth, I found that the development was poorly designed. There were attributes that were unnecessary when modifying the combination_id field in the calendar view. Right now, for the functionality I wanted to implement, it is only necessary to add the attribute filters="1" so that a list appears in the right-hand sidebar to filter by combination_id.

I also reviewed the other attributes since I was curious about what they actually do.

- **write_model:** Indicates the model to be modified (this model is solely for storing the filter information).
- **write_field:** Indicates the field to be modified (it must exist in the model specified in write_model).
- **filter_field:** Indicates whether it is active or not (it must exist in the model specified in write_model).
- **avatar_field:** Indicates the field where the image to be used is located (it must exist in the model specified in write_model).
 
In summary, these attributes provide the necessary information to retrieve the data for the filter widget.

@pedrobaeza

pedrobaeza:
Thanks for the extra information. Then please clean the PR only having the `filters="1"` and removing the rest of the changes.

adasatorres:
@pedrobaeza, Ready.

pedrobaeza (approved):
/ocabot merge minor

PR #142, opened 2024-12-26
[18.0][ADD] calendar_public_holiday

xaviedoanhduy:
### context:

- from this comment https://github.com/OCA/hr-holidays/pull/110#issuecomment-2096096367, extract some features from `hr_holidays_public` module into new module so that it can be used in other contexts and no longer depends on HR.

### note:

- rename the `hr.holidays.public` model and related models to `calendar.public.holiday` to match the Calendar repo.
- fix words like Holidays Public to Public holidays (because it is misspelled)
- moving tests and modifications from the old module will only concern the scope of functions in this module.


victoralmau (approved):
Code and functional review.

IMO the copyright of the corresponding files must be maintained to respect the attribution of all because this module is a “split” of the already existing `hr_holidays_public` module.

xaviedoanhduy:
> Code and functional review.
> 
> IMO the copyright of the corresponding files must be maintained to respect the attribution of all because this module is a “split” of the already existing `hr_holidays_public` module.

thanks for your suggestion, I have updated the copyright.

victoralmau (approved):
Ping @chienandalu

pedrobaeza:
Isn't the migration scripts missing here?

chienandalu:
> Isn't the migration scripts missing here?

I'd say that the hr_holidays_public stuff relative to time-off is goint to be still needed so the migration work could be done there, isn't it?

pedrobaeza:
I think it should be done here, as the public holidays should be transferred here.

xaviedoanhduy:
DRAFT: ADDING MIGRATION SCRIPT DATA

calendar_public_holiday/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "author": "Camptocamp," "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/calendar",
+    "depends": ["calendar"],
+    "data": [
+        "data/data.xml",
+        "security/ir.model.access.csv",
+        "views/calendar_public_holiday_view.xml",
+        "wizards/calendar_public_holiday_next_year_wizard.xml",
+    ],
+    "external_dependencies": {"python": ["openupgradelib"]},
```
victoralmau:
```suggestion
    "external_dependencies": {"python": ["openupgradelib"]},
```

I am not sure if this dependency should be added, what do you think @pedrobaeza  ?

pedrobaeza: No, it shouldn't be needed, unless it's used in business logic.

victoralmau:
Ping @pedrobaeza can you review it?

calendar_public_holiday/data/data.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--Copyright 2024 Camptocamp
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).-->
+<odoo>
+    <record id="event_type_holiday" model="calendar.event.type">
+        <field name="name">Holidays</field>
```
pedrobaeza: I think we should take the occasion to rename it:

```suggestion
        <field name="name">Public Holidays</field>
```

calendar_public_holiday/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    "name": "Calendar Holidays Public",
+    "summary": """
+        Manage Public Holidays
+    """,
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "category": "HR/Calendar",
+    "author": "Camptocamp," "Odoo Community Association (OCA)",
```
pedrobaeza: I think this module should have the same authors as the original `hr_holidays_public` one (apart from yours).

calendar_public_holiday/hooks.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2024 Camptocamp
```
pedrobaeza, line 1: Sorry for not noticing in my previous comment. As this is used in a hook, not on a migration script (that's what I meant for "business logic"), you should add the external dependency.

calendar_public_holiday/models/calendar_public_holiday.py
```diff
@@ -0,0 +1,126 @@
...
+import datetime
+
+from odoo import api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResourceCalendarPublicHoliday(models.Model):
+    _name = "calendar.public.holiday"
+    _description = "Calendar Public Holiday"
+    _rec_name = "year"
```
pedrobaeza, line 15: We should put first the most modern one.

```suggestion
    _rec_name = "year desc"
```

calendar_public_holiday/models/calendar_public_holiday.py
```diff
@@ -0,0 +1,126 @@
...
+    _order = "year"
+
+    year = fields.Integer(
+        "Calendar Year", required=True, default=fields.Date.today().year
```
pedrobaeza: If not, I think this is going to be a static value.

```suggestion
        "Calendar Year", required=True, default=lambda self: fields.Date.today().year
```

pedrobaeza (changes requested):
Thanks for the efforts. Apart from the inside comments, other points:

- A plain user shouldn't be able to create nor modify public holidays as it's happening right now (try with demo user in runboat).
- I think the menus should be inside Calendar > Configuration.

xaviedoanhduy:
> A plain user shouldn't be able to create nor modify public holidays as it's happening right now (try with demo user in runboat).

Hi @pedrobaeza, since this module no longer depends on `hr_holidays` but instead on the `calendar` module, so I changed the access of the models to be the same as in the dependent module (calendar) with the permission groups, see: https://github.com/odoo/odoo/blob/4209d441afa058a87e142b44d6e33a546741dd2b/addons/calendar/security/ir.model.access.csv?plain=1#L7-L8

pedrobaeza:
Yes, but they shouldn't follow the same rules, as public holidays are defined by managers, not by plain users. When adding `hr_holidays_public`, you can extend the permissions to allow HR managers to set them, but on this base module, only settings users should be able to modify this important information.

xaviedoanhduy:
> Yes, but they shouldn't follow the same rules, as public holidays are defined by managers, not by plain users. When adding `hr_holidays_public`, you can extend the permissions to allow HR managers to set them, but on this base module, only settings users should be able to modify this important information.

Ok, thanks for your advice and clarification. I have updated the permission group on the models of this module.

victoralmau (approved):

pedrobaeza:
Don't remove the super-menu "Public Holidays". Just host it inside the "Configuration" one. Please keep also the empty lines out.

pedrobaeza (approved):
Thank you!

/ocabot merge nobump

PR #164, opened 2025-07-01
[18.0][FIX] resource_booking: avoid importing Form outside of tests

### LINKED ISSUE #163: Issue with imports in resource_booking module
## Module
resource_booking

## Describe the bug
The Imported Form(from test) in controllers->portal.py file raises an error as 'Importing test framework, avoid importing from business modules and when not running in test mode' (can be found in the logs)

## To Reproduce
**Affected versions**:
18.0
1.Load the module

**Expected behavior**
The error should not come up. The use of Form should be avoided and replaced by some other mechanism

Image

carlos-lopez-tecnativa:
Since this commit https://github.com/odoo/odoo/commit/1f3a15a7d1f527ab05937810b6a1fb60c22ff455, when Odoo starts without tests enabled and the `Form` class is used in business code, an error appears in the log. This commit removes the usage of `Form`, which is not necessary anymore because no onchange methods depend on the start field, all fields are computed now.

Fixes https://github.com/OCA/calendar/issues/163

cc: @pedrobaeza @ByteMeAsap

### DIFF
--- a/resource_booking/controllers/portal.py
+++ b/resource_booking/controllers/portal.py
@@ -8,7 +8,6 @@ from dateutil.parser import isoparse
 
 from odoo.exceptions import AccessError, MissingError, ValidationError
 from odoo.http import request, route
-from odoo.tests import Form
 
 from odoo.addons.portal.controllers import portal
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Dependent computes should be triggered with this field assignation, so it seems OK.

Fast-tracking for greenifying the branch:

/ocabot merge patch

PR #168, opened 2025-08-04
[17.0][FIX] resource_booking: display all slots in the same time zone correctly

carlos-lopez-tecnativa:
When using a booking type with a resource calendar that has a different time zone than the resource itself, the slots are returned with mixed time zones. This commit normalizes the behavior and ensures that slots are always returned in the time zone of the booking type.

**Steps to reproduce:**

Create a resource with the time zone America/Guayaquil, and a calendar that starts from 06:00 to 15:00.

Create a resource booking type with a new calendar that starts from 09:00 to 16:00, and set the time zone to Europe/Madrid.

Before this commit, the slot displayed in the portal was 06:00, but it should have been 12:00. After this commit, the slot is displayed in the correct time zone (Madrid) at 12:00.

Note: The 06:00 in America/Guayaquil corresponds to 12:00 in Madrid standard time, not during daylight saving time. The tests use a freeze date in February.

Before
image

image

After
image
image

TT57463

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -7,6 +7,7 @@ import calendar
 from datetime import datetime, timedelta
 
 from dateutil.relativedelta import relativedelta
+from pytz import timezone
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Tested in real case.

victoralmau (approved):

PR #170, opened 2025-08-12
[18.0][FIX] resource_booking: change tag data to odoo

Saran440:


pedrobaeza:
Why is this needed? Can you check the CI?

Saran440:
I think `<data>` should change to `<odoo>` at all. But maybe it is not necessary to change.

PR #172, opened 2025-09-02
[15.0][FIX] resource_booking pre-migration script

sergiocorato:
A minimal fix.


pedrobaeza (approved):
Indeed, this is not having the openupgrade decorator, so it should be with `cr`.

pedrobaeza:
Let's merge it although the CI is red, as this is not influencing.

PR #188, opened 2025-10-27
[18.0][FIX] resource_booking: correct resource parameter in availability check

carlos-lopez-tecnativa:
The resources parameter must be a recordset, not a list. Before this commit, this was not a problem, but in this commit https://github.com/odoo/odoo/commit/79a559c9741410ad861c107e395b2fc486da95e8 , Odoo began expecting a recordset to call the function _get_calendar_att. This change was introduced during the migration to V15 in this commit https://github.com/OCA/calendar/commit/b7a6f5b2da40b8f3672966d5d93829f92ee8e7d8 , so we updated the parameter to pass it as a record.

### DIFF
--- a/resource_booking/models/resource_resource.py
+++ b/resource_booking/models/resource_resource.py
@@ -29,7 +29,7 @@ class ResourceResource(models.Model):
         # available here, we set the value to -1.
         result = self.calendar_id.with_context(
             analyzing_booking=-1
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Fast-tracking the patch for having it solved and not having red CIs:

/ocabot merge patch

PR #191, opened 2025-11-28
[18.0][ADD] caldav_sync: synchronize Odoo events with CalDAV servers

mdurepos:
New module (previously at git.bemade.org/bemade/bemade-addons, now being contributed to the OCA).

Permits two-way sync of calendar events between external CalDAV servers and Odoo calendar events.

PR #192, opened 2025-11-28
[19.0][ADD] caldav_sync: synchronize Odoo events with CalDAV server

mdurepos:
Allows the two-way synchronization of calendar events between Odoo and user-configured CalDAV servers.

Contributed to the OCA, originally on git.bemade.org/bemade/bemade-addons.


eicher31:
I tested and work fine!

ecino (approved):
👍 LGTM

XtremXpert (approved):
Working here

mdurepos:
Hello @pedrobaeza and @simahawk

Any chance that one of you can get this merged? I also have #193 and #191 pending in essentially the same state.

Thanks in advance!

caldav_sync/requirements.txt
```diff
@@ -0,0 +1,4 @@
+caldav>=1.3.9,<=2.0.1
```
pedrobaeza, line 1: This file should be out.

requirements.txt
```diff
@@ -0,0 +1,5 @@
+# generated from manifests external_dependencies
+caldav>=1.3.9,<=2.0.1
+icalendar
+markdown2
```
pedrobaeza, line 4: Why markdown is required in a caldav sync?

mdurepos: Because the description field on Odoo calendar events is HTML which renders incorrectly on most CalDAV calendar clients, so we convert it to/from markdown.

PR #199, opened 2026-01-14
[16.0][ADD] calendar_shared_ics

ntsirintanis:
This module allows sharing an Odoo calendar as a read-only ICS feed that can be
subscribed to in external calendar clients such as Outlook, Google Calendar,
and Thunderbird.

The calendar is exposed via a public URL protected by an access token and can be
added as a network calendar (ICS / ``webcal://``). External calendar clients periodically
refresh the feed, providing near-real-time visibility of Odoo events without any
write-back capability.


calendar_shared_ics/views/calendar_shared_ics_views.xml
```diff
@@ -0,0 +1,91 @@
...
+                        />
+                        <field
+                            name="domain"
+                            widget="domain"
+                            options="{'model': 'calendar.event'}"
+                        />
+                    </group>
+                    <group string="Subscription link">
+                        <field name="share_webcal_url" readonly="1" widget="copy" />
+                        <field name="share_url" readonly="1" widget="copy" />
```
Kiplangatdan: @ntsirintanis Correct widget should be `CopyClipboardChar`:
image

calendar_shared_ics/models/calendar_shared_ics.py
```diff
@@ -0,0 +1,225 @@
...
+        default=7,
+        help="How many days back to include for recent events",
+    )
+
+    def _compute_access_url(self):
+        """Adjust to modify access url"""
+        res = super()._compute_access_url()
+        base_url = self.get_base_url()
+        for cal in self:
+            cal.access_url = f"{base_url}/calendar/shared/{cal.id}/ics"
```
Kiplangatdan, line 64: @ntsirintanis Doesn't this need a portal view?

Kiplangatdan (changes requested):

calendar_shared_ics/models/calendar_shared_ics.py
ntsirintanis: Not really, this is just the url for that given calendar.shared.ics record, but without the access_token. It is just for being displayed

calendar_shared_ics/models/calendar_shared_ics.py
```diff
@@ -0,0 +1,225 @@
...
+            user_domain.append(
+                ("groups_id", "in", self.env.ref("base.group_portal").id)
+            )
+        # if both are ticked, add an OR
+        if len(user_domain) == 1:
+            users = self.env["res.users"].search(user_domain)
+        else:
+            users = self.env["res.users"].search(["|"] + user_domain)
+        partner_ids = users.mapped("partner_id").ids
+        return {"domain": {"partner_id": [("id", "in", partner_ids)]}}
```
Kiplangatdan: @ntsirintanis Two things here:

1. Returning a domain is deprecated in v16 in favor of using a computed function. 
E.g in [partner_contact_address_default](https://github.com/OCA/partner-contact/blob/16.0/partner_contact_address_default/models/res_partner.py#L24) and [usage in the view](https://github.com/OCA/partner-contact/blob/16.0/partner_contact_address_default/views/res_partner_views.xml#L15)

2. The current domain filter somehow doesn't work well when you include portal users:
image

Kiplangatdan (changes requested):

calendar_shared_ics/models/calendar_shared_ics.py
Kiplangatdan: image

Kiplangatdan (approved):
@ntsirintanis Retested, LGTM!

PR #200, opened 2026-01-21
[18.0] [IMP] resource_booking: handle partner_ids (booking attendees) safely 

yessser:
following same logic as #155


pedrobaeza (approved):
/ocabot merge patch

PR #210, opened 2026-03-16
[18.0][FIX] resource_booking: Add groups to calendar event form view

victoralmau:
FWP from 17.0: https://github.com/OCA/calendar/pull/212

When a user had no resource_booking permissions and opened a calendar event linked to a resource booking, he got an unnecessary access error.

**Before**:
ejemplo

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa

### DIFF
--- a/resource_booking/tests/test_backend.py
+++ b/resource_booking/tests/test_backend.py
@@ -37,6 +37,19 @@ class BackendCaseBase(TransactionCase):
 @freeze_time("2021-02-26 09:00:00", tick=True)  # Last Friday of February
 class BackendCaseMisc(BackendCaseBase):
     @users("plain")
+    def test_plain_user_calendar_event_form(self):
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Is the invisible field needed in 18?

victoralmau:
> Is the invisible field needed in 18?

I'm not 100% sure that it's necessary to display the field, but I'd rather avoid any unintended consequences if the field is removed and doesn't get automatically added back or doesn't retain certain functionality. Therefore, I prefer to keep it but add the “groups” field (just as it has been added in other versions).

pedrobaeza:
You can just try the same test removing the field and see if it fails.

pedrobaeza (approved):
/ocabot merge patch

PR #214, opened 2026-05-13
[18.0][FIX] resource_booking: Add description to meeting values only when the field is set

carlos-lopez-tecnativa:
For a curious reason, the `test_resource_booking_schedule_unschedule` test fails after this PR: https://github.com/odoo/odoo/pull/223875 The `requester_advice` field is of type `Text`, but in `calendar.event` the `description` field is `HTML`. I tried changing the field type to HTML, but the error still occurs, so another approach is to only pass the field when it is filled.

See log error in https://github.com/OCA/calendar/pull/213


pedrobaeza:
I did #215 and merge for solving the template issue. Rebased over it.

pedrobaeza (approved):
Now it's working! Weird side effect...

/ocabot merge patch

PR #187, opened 2025-10-19
[19.0][MIG] resource_booking: Migration to 19.0

tishmen:
Scope
- resource_booking only

Summary
- Migrate to Odoo 19 using the working feature branch code.
- Context handling aligned with 19.0.
- Asset paths are relative under    and frontend bundles.
- Temporary test dependency to web addon referenced via PR below.

Pre-commit
- Verified locally; no additional changes applied here to keep parity with the feature branch.

Temporary dependency
- Added a temporary reference in 
  
  
  
  
  
  
  
  
   (root) to install from the PR ref, per OCA guidelines:

Tests
Command:

Result: 45 tests green locally.

Depends on:
- [ ] https://github.com/OCA/web/pull/3306

PR #176, opened 2025-10-03
[18.0][FIX] resource_booking: add display_name in test

carlos-lopez-tecnativa:
Odoo added this new key in commit
https://github.com/odoo/odoo/commit/c840bdedb99320360b31dc3674022ea01800832a.

To prevent the test from failing, we updated the data accordingly.

image

https://github.com/OCA/calendar/actions/runs/17429589563/job/49484640129#step:8:449

@Tecnativa @pedrobaeza @victoralmau could you please review this?

PR #217, opened 2026-05-16
[19.0][MIG] resource_booking: Migration to 19.0

dnplkndll:
Port of `resource_booking` from 18.0 to 19.0. Follows the [OCA migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0).

Supersedes #187 (open since 2025-10-19, original author unreachable).

## Non-mechanical adaptations worth flagging

- `Intervals` moved from `odoo.addons.resource.models.utils` → `odoo.tools.intervals`.
- `_sql_constraints` list rewritten as `models.Constraint(...)` declarations (Odoo 19 SQL constraint API).
- Portal controller drops `datetime.utcfromtimestamp` (deprecated in Python 3.12) for `datetime.fromtimestamp(..., tz=timezone.utc).replace(tzinfo=None)`.
- Portal template's `t-field="meeting.display_time"` switched to `t-out` calling a new `booking._get_portal_display_time()` helper — Odoo 19 dropped the computed `calendar.event.display_time` field.
- Tour step in `resource_booking_tour.esm.js` annotated with `expectUnloadPage: true` per Odoo 19's tour framework's new contract for click-triggers that navigate.

## 18.0 fixes forward-ported into [MIG]

The cherry-pick replay included [FIX] commits that the [MIG] commit accidentally dropped (written against an older baseline). All four are now in [MIG]:

- [#183](https://github.com/OCA/calendar/pull/183): restore `_get_intervals(tz)` + TZ-aware slot conversion so portal slots always display in the booking type's time zone. `WorkIntervals` (removed in 19.0 with `hr_work_entry_contract`) replaced by `Intervals`.
- [#189](https://github.com/OCA/calendar/pull/189): inject `event_tz` from the resource calendar on meeting **creation** only (cannot be changed on write) so the calendar event carries the correct time zone from the start.
- [#214](https://github.com/OCA/calendar/pull/214): set `description` on the meeting only when `requester_advice` is filled — unconditional assignment caused `test_resource_booking_schedule_unschedule` to fail after odoo/odoo#223875 (Text → HTML field type mismatch on write).
- [#219](https://github.com/OCA/calendar/pull/219): honor all-day calendar events when computing busy slots. `expression.OR` (removed in 19.0) replaced by `Domain.OR` from `odoo.fields`; interval rendered from `start_date`/`stop_date` in the analyzer's tz.


victoralmau:
Please, cherry-pick https://github.com/OCA/calendar/pull/222 to commit history before migration commit.

dnplkndll:
Done — cherry-picked both commits from #222 (`0a4d552` [FIX] + `64d0441` [IMP]) in front of the migration commit, also rebased onto current upstream/19.0 to absorb #223 (copier-template 1.42 update).

Side-effect note: my original `[FIX] checklog: ignore chrome-zombie warnings…` commit dropped during the rebase — upstream now has a broader `WARNING .* Killing chrome descendants-or-self .*` pattern in `checklog-odoo.cfg` that supersedes the narrower one I had, so the commit was redundant. Net result: 7 commits ahead of `19.0` instead of 8.

CI re-running on the force-push. Branch: https://github.com/ledoent/calendar/tree/19.0-mig-resource_booking

pedrobaeza:
The commit history is incomplete (it's not from the beginning), and now that you are rebuilding it, you can include the merging PR #219

dnplkndll:
@pedrobaeza @victoralmau 

ready for review

resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,830 @@
...
+            record.user_id = record.meeting_id.user_id
+
+    def _prepare_meeting_vals(self):
+        resource_partners = self.combination_id.resource_ids.filtered(
+            lambda res: res.resource_type == "user"
+        ).mapped("user_id.partner_id")
+        return dict(
+            alarm_ids=[(6, 0, self.type_id.alarm_ids.ids)],
+            categ_ids=[(6, 0, self.categ_ids.ids)],
+            description=self.type_id.requester_advice,
```
carlos-lopez-tecnativa: In this commit cd7ebb4 you included the changes from this PR #214

But in this commit f8c5ec4 you reverted those changes. Why?

resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,830 @@
...
+                            one.meeting_id.start == one.start,
+                            one.meeting_id.stop == one.stop,
+                            one.meeting_id.duration == one.duration,
+                        )
+                    ):
+                        # Context to notify scheduling change
+                        meeting = meeting.with_context(from_ui=True)
+                    meeting.write(meeting_vals)
+                else:
+                    to_create.append(meeting_vals)
```
carlos-lopez-tecnativa, line 424: In this commit https://github.com/OCA/calendar/pull/217/changes/b7becb8168473c31532f56fb9eac3f19e2acb1f3 you included the changes from this PR https://github.com/OCA/calendar/pull/189 

But in this commit https://github.com/OCA/calendar/pull/217/changes/f8c5ec4b5a08301228f612836ba85be158220f70 you reverted those changes. Why?

resource_booking/models/resource_booking_combination.py
```diff
@@ -0,0 +1,119 @@
...
+                if not combination_intervals:
+                    break  # Can't restrict more
+                calendar = combination.forced_calendar_id or res.calendar_id
+                # combination_intervals &= calendar._work_intervals(start_dt,
+                # end_dt, res)
+                combination_intervals &= calendar._work_intervals_batch(
+                    start_dt, end_dt, res
+                )[res.id]
+            result |= combination_intervals
+        return result
```
carlos-lopez-tecnativa, line 107: In this commit https://github.com/OCA/calendar/pull/217/changes/10d55b91d11be0d6454f68a4803fba338b79b661 you included the changes from this PR https://github.com/OCA/calendar/pull/183 

But in this commit https://github.com/OCA/calendar/pull/217/changes/f8c5ec4b5a08301228f612836ba85be158220f70 you reverted those changes. Why?

carlos-lopez-tecnativa (changes requested):
Please check whether it is necessary to include this PR https://github.com/OCA/calendar/pull/225

Additionally, please squash the administrative commits, such as those shown in the screenshot.

Image

resource_booking/tests/test_backend.py
```diff
@@ -507,37 +507,6 @@ def test_recurring_event(self):
...
-        rb_f.type_id = self.rbt
-        # Force the user-resource combination so the all-day event has to block it
-        rb_f.combination_auto_assign = False
-        rb_f.combination_id = self.rbcs[0]
-        rb_f.start = datetime(2021, 3, 1, 9)
-        with self.assertRaises(ValidationError):
-            rb_f.save()
-        # Following Monday is fine
-        rb_f.start = datetime(2021, 3, 8, 9)
-        rb_f.save()
```
carlos-lopez-tecnativa: Why was the test added in this commit dropped?
https://github.com/OCA/calendar/pull/217/changes/3cac0872b9b59f3edbda5498e7226303f6e71e65

resource_booking/tests/test_backend.py
```diff
@@ -978,50 +943,6 @@ def test_resource_is_available_span_days(self):
...
-            }
-        )
-        response = resource_booking._get_calendar_context()
-        for slots in response["slots"].values():
-            # The first slot should start at 06:00 in America/Guayaquil.
-            # In Europe/Madrid, it should be 12:00.
-            self.assertEqual(slots[0].strftime("%H:%M:%S"), "12:00:00")
-            # Check that all slots are in the expected time zone.
-            for slot in slots:
-                self.assertEqual(slot.tzinfo.zone, "Europe/Madrid")
```
carlos-lopez-tecnativa: Why was the test added in this commit dropped?
https://github.com/OCA/calendar/pull/217/changes/3c084be5a501b6ba4ccd6821ec18e9202b62079a

carlos-lopez-tecnativa (changes requested):
If you removed the test, please add a justification to explain the reason for the change. Or was it dropped by mistake? If so, please restore it in the corresponding commit.

resource_booking/tests/test_backend.py
dnplkndll: error in the squash, replaced

resource_booking/tests/test_backend.py
dnplkndll: error in the squash, replaced

resource_booking/models/resource_booking.py
dnplkndll: reworked

resource_booking/models/resource_booking.py
dnplkndll: reworked to preserve

resource_booking/models/resource_booking_combination.py
dnplkndll: reworked to preserve

resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,837 @@
...
+        unfitting_bookings = has_meeting
+        now = fields.Datetime.now()
+        for booking in has_meeting:
+            # Ignore if the event already happened
+            already_happened = booking.stop and booking.stop < now
+            if already_happened:
+                unfitting_bookings -= booking
+                continue
+            start_dt = fields.Datetime.context_timestamp(self, booking["start"])
+            end_dt = fields.Datetime.context_timestamp(self, booking["stop"])
```
carlos-lopez-tecnativa, line 468: When I have a meeting and try to remove the dates, I get this error. Note that in v18, performing the same action does not trigger the error. Please take a look.

Image

carlos-lopez-tecnativa (changes requested):

resource_booking/models/resource_booking.py
```diff
@@ -518,12 +518,67 @@ def _get_calendar_context(self, year=None, month=None, now=None):
...
+        stop_dt = fields.Datetime.context_timestamp(
+            self_tz, fields.Datetime.from_string(meeting.stop)
+        )
+        # Date from current language; time forced to 24-hour
+        date_fmt = get_lang(self.env).date_format
+        time_fmt = "%H:%M:%S"
+        date_str = start_dt.strftime(date_fmt)
+        start_str = start_dt.strftime(time_fmt)
+        end_str = stop_dt.strftime(time_fmt)
+        return f"{date_str} at ({start_str} To {end_str}) ({timezone})"
```
carlos-lopez-tecnativa: I'm not sure whether these changes, which I consider to be a new feature, should be merged as part of the migration process. I think it would be better to complete the migration first and then open a new PR to address this functionality.
What do you think, @pedrobaeza and @victoralmau?

resource_booking/models/resource_booking.py
```diff
@@ -669,35 +724,65 @@ def _message_auto_subscribe_followers(self, updated_values, default_subtype_ids)
...
+        return [
+            {
+                "lang": None,
+                "partner_id": p.id,
+                "name": p.name,
+                "display_name": p.display_name,
+                "reason": reason,
+            }
+            for p in self.partner_ids
+        ]
```
carlos-lopez-tecnativa: I'm not sure about this change. Could you please explain in more detail what issue occurs without it? I remember a similar case being handled in this PR: https://github.com/OCA/calendar/pull/176.

I also don't like that the super() method is not called. Instead of completely replacing the logic, wouldn't it be better to call super() and then adapt the result according to the new requirements? This would preserve the standard behavior while applying only the necessary changes.

carlos-lopez-tecnativa (approved):
Functional review: LGTM.

I left two comments about the changes, but if others think this can be included as part of the migration, I'm fine with moving forward.

TT58697

resource_booking/models/resource_booking.py
dnplkndll: ok, yes I agree. better to make the test match the new formatting.

resource_booking/models/resource_booking.py
dnplkndll: agreed. now calling super and appending the partner_ids to the results to preserver the form.

joachimageron:
Hi @dnplkndll, thanks a lot for working on this migration! 🙏

This module is really valuable for Odoo 19 Community users, since the native
`appointment` app is now Enterprise-only, so a lot of people are looking forward
to having `resource_booking` available on 19.0.

The PR looks in great shape — it's mergeable with no conflicts and already has an
approval from @carlos-lopez-tecnativa. It seems to only be waiting for a second
review to be merged.

Could a second maintainer take a look when they get a chance? Thanks a lot for the
great work! 🙌

pedrobaeza:
@joachimageron you can review it as well

thierry-cmd (approved):
Functional review on Odoo 19.0 Community — LGTM.

### Test setup

- Odoo 19.0 Community from source, Python 3.12, PostgreSQL 18, Windows.
- Fresh database, no other custom module.
- This PR's branch (`ledoent:19.0-mig-resource_booking`) installed together with
  `website_appointment_booking` (19.0, same author) in order to exercise the public path.
- Config: one working schedule (timezone Europe/Brussels, with a `lunch` period), one human
  resource, one combination, one booking type (duration 1:00, slot duration 1:15 then 1:30,
  modifications deadline 24:00).
- Real SMTP configured; public flow played in a private window; invitation e-mail read in a
  real mailbox.

### What works

- Install and upgrade are clean, no traceback.
- Booking type / combination / availability calendar configuration behaves as documented.
- Slot computation is correct: starts are anchored at the beginning of each attendance
  interval and spaced by `slot_duration`; a slot is only offered if `duration` fits entirely
  in the interval; `day_period = 'lunch'` intervals are never entered; an entry in
  `resource_calendar_leaves` removes the whole day from the public page.
- Timezone handling in the booking itself is correct: `calendar_event.start` stored as
  11:00 UTC for 13:00 Europe/Brussels (DST), matching the working schedule timezone.
- `calendar.event` is created with consistent `start`/`stop`/`duration`, `allday = False`.
  `res_model`/`res_id` are empty by design (`_prepare_meeting_vals`); the link is kept
  through `resource_booking.meeting_id`, so the event is not orphaned.
- Concurrency: two private windows booking the same slot a few seconds apart — the first one
  confirms, the second gets "That slot is no longer available. Please choose another." and is
  sent back to the calendar. No crash, no double booking.
- `modifications_deadline` correctly excludes the next 24h from the offered slots
  (`start_dt = max(start_dt, now + modifications_deadline)`).
- The portal page itself is good: full recap, resource, dates, advice, public chatter,
  Feedback and Cancel buttons. `_get_booking_sudo()` correctly forces the resource calendar
  timezone on the portal context.
- Cancelling is effective: `state = canceled`, `active = False`, meeting unlinked.

### Finding 1 — a timezone fix from 18.0 is silently dead in 19.0

`models/calendar_event.py` still overrides `get_interval()`, whose docstring states that any
notification related to a booking must be emitted in the resource booking timezone. That
method no longer exists in `calendar/models/calendar_event.py` in 19.0 — the invitation
template now resolves the timezone from `calendar.attendee.mail_tz`, which is computed as
`partner_id.tz` only. The override is therefore never called, and its `super()` would target
a missing method.

Consequence: invitation e-mails for a booking are rendered with `tz=False` whenever the
attendee's partner has no timezone, so the time is printed in UTC — and since the template
guards the timezone label with `t-if="object.mail_tz"`, no timezone is shown at all. Measured:
a booking at 13:00 Europe/Brussels produced an e-mail body reading `11:00 AM` with no timezone
mention, while the attached `invitation.ics` was correct. A customer reading the text would
arrive two hours early; a customer clicking "add to calendar" would not.

Note this is not an edge case: partners created by the public booking route are created with
name and e-mail only, so they never have a timezone.

A minimal way to restore the 18.0 behaviour, independently of partner data:

```python
class CalendarAttendee(models.Model):
    _inherit = "calendar.attendee"

    def _compute_mail_tz(self):
        super()._compute_mail_tz()
        for attendee in self:
            booking = attendee.event_id.sudo().resource_booking_ids
            booking_tz = booking.type_id.resource_calendar_id.tz
            if booking_tz:
                attendee.mail_tz = booking_tz
```

and dropping the now-dead `get_interval()` override.

I applied this on top of your branch and replayed a public booking: the invitation body then
reads `Thursday 13 August 2026 01:00 PM (Europe/Brussels)` instead of `11:00 AM` with no
timezone, with the attendee's partner still having no timezone set.

### Other findings

These are not migration regressions as far as I can tell — they look like pre-existing
behaviour. Reported for follow-up, they did not block my testing.

**2. Declining the calendar invitation silently keeps the slot booked.**
`_compute_state` derives `canceled` from `active` only, and `active` is only changed by
`action_cancel()`. Declining sets `attendee.state = 'declined'`, so `confirmed` becomes False
and the state falls back to `scheduled`. Since a booking made through the public route marks
the requester as `accepted`, declining produces a silent `confirmed -> scheduled` transition.
Measured on four bookings of the same day: three `Confirmed`, the declined one back to
`Scheduled`, with the slot still held. From the backend that state is indistinguishable from
"has not answered yet", and the customer sees a confirmation page after declining.

**3. Portal cancel redirects public users to an authenticated page.**
`portal_booking_cancel()` is declared `auth="public"` but ends with `request.redirect("/my")`,
which requires a login. `action_cancel()` has also cleared the `access_token`, so the user's
link is dead by then. The cancellation does take place, but the user lands on a login form
with no confirmation of it. A redirect to a public confirmation page would be enough.

**4. `access_token` is not generated for bookings created through a public flow.**
`_portal_ensure_token()` only runs when a share URL is requested (portal mail from the
backend, or the Preview button). Bookings created by a public route keep an empty
`access_token` — verified in SQL on three bookings created from the website. The portal page,
which is good, is therefore unreachable unless a backend user shares it explicitly.

Thanks for the work on this migration.

PR #219, opened 2026-05-16
[18.0][FIX] resource_booking: honor all-day calendar events when computing slots

dnplkndll:
All-day calendar events store dates in `start_date`/`stop_date` and have no `start`/`stop` datetimes, so the busy-events query in `_calendar_event_busy_intervals` ignored them. A user's all-day event (PTO, off-site) silently failed to block bookings on that day.

This change OR-extends the conflicting-events search with an allday + `start_date`/`stop_date` predicate, and renders the busy interval from `start_date` through `stop_date + 1 day` in the analyzer's tz so a single all-day event blocks the full local day across tz boundaries.

Adds `test_allday_event_blocks_booking_slot` covering the regression.

`/ocabot merge patch`

### DIFF
--- a/resource_booking/models/resource_calendar.py
+++ b/resource_booking/models/resource_calendar.py
@@ -2,9 +2,12 @@
 # Copyright 2022 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from datetime import datetime, time, timedelta
+
 from pytz import UTC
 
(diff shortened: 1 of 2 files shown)


victoralmau:
Ping @carlos-lopez-tecnativa

pedrobaeza (approved):
Tested on real case where previously where showing slots with all-day busy events.

carlos-lopez-tecnativa (approved):

PR #220, opened 2026-05-16
[18.0][IMP] resource_booking: cap suggested slots with max_advance_booking_days

dnplkndll:
Add an Integer field `max_advance_booking_days` on `resource.booking.type` (SQL-constrained nonnegative, default 0) and gate `_get_available_slots` to skip starts beyond `now + N days`. Default 0 preserves the existing unbounded behavior.

Useful when a booking type should not let requesters reserve weeks or months out (office desks released only one week ahead, sales-call windows that should not commit dates beyond a planning horizon).

Adds 3 tests in `tests/test_backend.py`: `test_max_advance_booking_days_caps_slots`, `..._zero_keeps_full_range`, `..._constraint`.

`/ocabot merge minor`

### DIFF
--- a/resource_booking/models/resource_booking.py
+++ b/resource_booking/models/resource_booking.py
@@ -561,6 +561,13 @@ class ResourceBooking(models.Model):
         start_dt = max(
             start_dt, now + timedelta(hours=self.type_id.modifications_deadline)
         )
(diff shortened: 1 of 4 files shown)


brenden-eshbach:
Thanks for your work on this @dnplkndll

